PR #4, opened 2023-02-19
Fix minor misspelled text in UI

astronaut701:
Hey, I noticed that some text was mistyped and corrected it for you. Feel free to check it over one more time to make sure I'm not making a mistake too.

Also, kudos for figuring all this out!

### DIFF
--- a/Settings.cs
+++ b/Settings.cs
@@ -117,7 +117,7 @@ namespace GHelper
             {
                 button60Hz.Enabled = false;
                 button120Hz.Enabled = false;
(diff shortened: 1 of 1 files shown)


astronaut701:
Oh, I'd also suggest posting a screenshot with the corrected text and options whenever you get a chance too. No rush, just a reminder from me.

seerge:
Thanks :) i was rushing so much, so didn't notice my typos. Unfortunately visual studio doesn't seem to do a spellcheck.

PR #7, opened 2023-02-19
Assure gpu mode and refresh rate are set at startup

Albert24GG:


seerge:
To be honest, I'm not sure if it's a good user experience, when on app launch it will suddenly (without any user action) shut down GPU for example :| I deliberately don't apply GPU mode on startup (even tho i know it from settings) as this is the most invasive action that app can do.

Albert24GG:
It only takes effect when the "Set Eco on battery and Standard when plugged" and  "Set 60Hz on battery, and back when plugged" options are checked, in order to do what it's meant to do. This is also the behavior in Armoury Crate, which sets eco on boot if I have that option set, so I would not consider it invasive.

Albert24GG:
For example, if I shutdown with the laptop being plugged in and the gpu mode set on standard, when I power on the laptop the gpu mode will not automatically get set to eco and the refresh rate will remain at max. This means that I will have to manually adjust the gpu mode and refresh rate, which defeats the purpose of the feature.

astronaut701:
I'll concur with @Albert24GG here. Having to adjust everything when I just started up the laptop away from the charger creates friction with the program's use, in my opinion. @seerge, if you really feel strongly about this not being a good idea, perhaps make this option an additional toggle that explicitly says what it's going to do as a compromise should you want to implement this.

seerge (commented):
Assure gpu mode and refresh rate are set at startup

PR #9, opened 2023-02-20
Added option to switch between silent/balanced mode when on battery/plugged

Albert24GG:
I thought that a good addition to the other options would be to automatically switch to silent mode when running on battery, since this could have a positive impact on battery life. Something that can be improved is the placement of the checkbox, since I couldn't get it working without breaking the other objects around.
Screenshot 2023-02-20 234630

### DIFF
--- a/GHelper.csproj
+++ b/GHelper.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net7.0-windows</TargetFramework>
+    <TargetFramework>net7.0-windows10.0.17763.0</TargetFramework>
     <Nullable>enable</Nullable>
     <UseWindowsForms>True</UseWindowsForms>
     <ImplicitUsings>enable</ImplicitUsings>
@@ -31,6 +31,7 @@
 
   <ItemGroup>
     <PackageReference Include="hidlibrary" Version="3.3.40" />
+    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications" Version="7.1.3" />
     <PackageReference Include="System.Management" Version="7.0.0" />
     <PackageReference Include="TaskScheduler" Version="2.10.1" />
   </ItemGroup>
--- a/Program.cs
+++ b/Program.cs
@@ -531,12 +531,14 @@ namespace GHelper
 
             settingsForm.VisualiseGPUAuto(config.getConfig("gpu_auto"));
             settingsForm.VisualiseScreenAuto(config.getConfig("screen_auto"));
+            settingsForm.VisualiseCPUAuto(config.getConfig("cpu_auto"));
 
             settingsForm.SetStartupCheck(scheduler.IsScheduled());
 
             bool isPlugged = (SystemInformation.PowerStatus.PowerLineStatus == PowerLineStatus.Online);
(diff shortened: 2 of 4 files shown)


seerge:
It's possible and I have thought about it before.But do people really need it ? 

Silent mode is NOT more economical than Balanced (unlike Eco GPU mode or low refresh rate), it's just more limited in terms of TDP, that's it. I easily get 6-7w battery drain on Balanced when do browsing or so.

My main concern here is not to overload app with too many options :D or I will have to do some extra settings page for them.

Albert24GG:
Silent mode should also allow higher temperatures with no fans spinning, so that would help with battery drain a little in some scenarios.

masi0:
> It's possible and I have thought about it before.But do people really need it ?
> 
> Silent mode is NOT more economical than Balanced (unlike Eco GPU mode or low refresh rate), it's just more limited in terms of TDP, that's it. I easily get 6-7w battery drain on Balanced when do browsing or so.
> 
> My main concern here is not to overload app with too many options :D or I will have to do some extra settings page for them.
> 
> Up to discussion :)

if auto profile switching = avoid additional application to be install then it could be a great and only alternative to armored crate!

astronaut701:
Why not have an option to auto switch profiles that were selected when last plugged/last unplugged. Isn't that how Armory Crate does it?

flemesre:
I think in this case it might make more sense to change the option to disabling CPU boost on battery for those who want it. Like @seerge said, the performance modes only really changes TDP, and I haven't noticed much difference in power draw between Silent and Balanced on my G15 2021. Disabling turbo boost, however, makes a big difference (~20W bursts to ~9W bursts) in peak CPU draw according to HWInfo64.

yummypugs:
I just started using this app, and I would really appreciate an option like this. I like to use my laptop with 0 fans spinning when I'm in University and perhaps there is some battery saving to be found, as the computer is allowed to run hotter without spinning up fans.

seerge:
But CPU boost checkbox is already in app ? So boost can be disabled? Or I'm not getting something

flemesre:
> But CPU boost checkbox is already in app ? So boost can be disabled? Or I'm not getting something

I meant adding an option to disable the boost automatically when switching from plugged in to battery, similar to the GPU mode and refresh rate switched that are already implemented.

seerge:
> > But CPU boost checkbox is already in app ? So boost can be disabled? Or I'm not getting something
> 
> I meant adding an option to disable the boost automatically when switching from plugged in to battery, similar to the GPU mode and refresh rate switches that are already implemented.

CPU Boost is part of windows power plan, and it's aready 2 separate settings for on / off battery

https://imgur.com/a/ZaAoGPB

ke1ne:
> It's possible and I have thought about it before.But do people really need it ?

Yeap, this could be nice.

seerge:
this feature is already implemented, app remembers states on battery/power

seerge:
Sorry, for closing a pull request btw, i thought that I'm answering to an issue. But anyway, current logic of performance modes is different - as app remembers them per battery/power state, so switch is not needed :)

Albert24GG:
> Sorry, for closing a pull request btw, i thought that I'm answering to an issue. But anyway, current logic of performance modes is different - as app remembers them per battery/power state, so switch is not needed :)

No problem!:) It's nice you implemented a solution for that.

PR #14, opened 2023-02-21
Send notification on performance mode cycling

Albert24GG:
I made the program send a notification whenever performance mode is changed through the keybinding and I set it to be the highest priority. Still, to appear over fullscreen games, one has to allow GHelper in Priority Notification Settings, in Windows Settings, because there is no other way to show the notification over a fullscreen app.


seerge:
Looks cool, I have tested it :) I also tried native notifications before. 
But main problem that you don't see it in game, and that's probably main scenario of using a hotkey.
What I wanted to do - is same notifications as armory crate was doing.

I have asked about that on stackoverflow https://stackoverflow.com/questions/75514689/c-sharp-most-simple-way-to-show-small-overlay-similar-to-windows-11-brightness-v but without success yet ...

seerge:
Oke, let's try it like that. May be eventually I will figure out the way armory was showing it's own notifications :)

Albert24GG:
@seerge I think the program can be added to that priority list editing some registry key. Perhaps that's what armoury crate does. I will try to see if that works.

seerge:
Have you seen armory's notifications ? They are custom, and appear in center :)

seerge:
Btw, we have a problem, if i set target OS to windows 10 (to support this native toasts), it tries to embed 20MB of Microsoft.Windows.SDK.NET.dll into exe making it 10 times bigger than it was :|

seerge:
and notifications are still not visible above games

Albert24GG:
> Have you seen armory's notifications ? They are custom, and appear in center :)

Oh, I remember them. Idk if we can achieve that, but I will try to research.

> Btw, we have a problem, if i set target OS to windows 10 (to support this native toasts), it tries to embed 20MB of Microsoft.Windows.SDK.NET.dll into exe making it 10 times bigger than it was :|

Unfortunately, without windows 10 target os, native notifications won't work://

astronaut701:
> Have you seen armory's notifications ? They are custom, and appear in center :)

If I'm not mistaken, those notifications (as well as other notifications for keyboard, trackpad off, etc.) are run by an executable called AsusOSD.exe

It's stored under C:\Windows\System32\DriverStore\FileRepository\asussci2.inf_amd64_0100494bef227dd5\ASUSOptimization for me.

I might take a shot in the dark of trying to see how it works and poking it into giving us the notifications Armory Crate usually gives us after I'm finished with dark mode.

seerge:
I have stashed this changes into separate branch. That toast requires win10 build in settings, and it increases file size from 2mb (which is already a lot) to 23(!) mb cause it includes .net library there. And it seems that without that lib it won't work ...

Albert24GG:
I will try to redesign the notifications, perhaps make it pop up on top of fullscreen programs.

PR #17, opened 2023-02-23
Added toast notification when switching performance mode

Albert24GG:
I saw that you started to make a new implementation for the notification so I decided to try to finish it. The notification appears in the middle of the screen, on top of other windows. It works with most fullscreen games as well. I tested it on CSGO, MS Flight Simulator and Rocket League, and only on Rocket League it seemed to not work(I think the game uses some sort of protection or something). I also used the white version of the icon since it looks better imo.
Screenshot 2023-02-20 234630


seerge:
Yeah .. but if you run game not on native fullscreen resolution (my use case is battlefield 2042 in 1440p on 4K monitor with RSR) it just kicks me out of the game completely (by minimizing game). That's why i didn't go with that way :|

Albert24GG:
> Yeah .. but if you run game not on native fullscreen resolution (my use case is battlefield 2042 in 1440p on 4K monitor with RSR) it just kicks me out of the game completely (by minimizing game). That's why i didn't go with that way :|

Have you tried running the game in borderless windowed?

seerge:
Borderless window has issues with RSR. But in general my point was that such toast doesn't work :D

Albert24GG:
Well then idk what else would work. Doing it with a directx overlay seems complicated and risky...

seerge:
Yeah ... direct x overlay is something cheaters use, and is a grey area. I'm just curious how armory was doing that ...

astronaut701:
I still need to dig into how AsusOSD.exe works and what triggers it. That may lead us down a path. I posted elsewhere where it could be found, but can't seem to find the post right now.

seerge:
After few hours of suffering, I have found some working example and turned it into working OSD notification, that doesn't kick me out of games and seems to work.

It works in all DirectX 12 games (where fullscreen is not exclusive), but not in DX11-- tho. But that's at least something.

Albert24GG:
@seerge I analyzed a bit the AsusOSD.exe with ghidra and they seem to be using the win32 api since there are some calls of ShowWindow(), SetForegroundWindow() and some other.The code is written in C though, but I think this shouldn't matter.

seerge:
Well, that's pretty much same as current implementation does then :)

Albert24GG:
Then how does it work better if not in fullscreen exclusive? This toast was also working in fullscreen in flight sim 2020, which is directx 12. I don't get how this is more useful ://

seerge:
that implementation doesn't kick out of fullscreen games, at least i didn't manage to reproduce it

Albert24GG:
> that implementation doesn't kick out of fullscreen games, at least i didn't manage to reproduce it

Without one or two added options, it behaves exactly the same as this one, but I see...

seerge:
I don't know exact reason why it didn't work in your branch, but i deliberately pulled it again, compiled - and it doesn't work on battlefield 2042 for example :| may some flags, ... 

All this ancient win32 apis are horrible. Last time (before GHelper) i did something for windows - was in times of windows xp and in school :D. And some things seem to be as bad as they were ... :D

Albert24GG:
Yeah, If armoury crate had been open source, it would have been so much easier, but it remains a mistery how they do it:]]

PR #74, opened 2023-03-13
feat: added GPU temperature indication. 

ZimM-LostPolygon:
- Supports both NVIDIA and AMD discrete GPUs.
- AMD implementation is based on https://github.com/GPUOpen-LibrariesAndSDKs/display-library, but heavily changed to support ADL2. It worked for me against the integrated GPU in G14 2021, but some testing on the 2022 model with AMD is required. Specifically, distinguishing between integrated and discrete GPUs:
https://github.com/ZimM-LostPolygon/g-helper/blob/62ac478761ffbe1e6dc200fc622950342dc3fda1/Gpu/AmdGpuTemperatureProvider.cs#L50
- Immediately update sensors when opening GHelper window. Before, I had to wait a few seconds for the first value...

### DIFF
--- a/GHelper.csproj
+++ b/GHelper.csproj
@@ -39,6 +39,7 @@
   <ItemGroup>
     <PackageReference Include="hidlibrary" Version="3.3.40" />
     <PackageReference Include="HidSharpCore" Version="1.2.1.1" />
+    <PackageReference Include="NvAPIWrapper.Net" Version="0.8.1.101" />
     <PackageReference Include="System.Management" Version="7.0.0" />
     <PackageReference Include="TaskScheduler" Version="2.10.1" />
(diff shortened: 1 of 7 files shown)


seerge:
The only thing that worries me, that on every timer-refresh (when main GHelper window is visible) it will check for both GPU Providers every time, it may cost some extra resources :)

ZimM-LostPolygon:
It will not :) The providers are refreshed only on events that trigger hardware change. Once it's created, only that one provider will be used for timer-based refresh

seerge:
Welcome to my world @ZimM-LostPolygon :)
1. https://www.reddit.com/r/ZephyrusG14/comments/10z6nhw/comment/jc3rs9r/?context=3
2. https://github.com/seerge/g-helper/issues/81 (most probably GPU sensor)

PR #97, opened 2023-03-18
Turkish Translation

HamzaYslmn:
I really like the app, thank you. I translated for Turkish users. I localized the usage and interface.


HamzaYslmn:
I add the dark mode icons

seerge:
That's cool, that you put this efforts, but currently there is no multilingual support in app (and won't be in foreseeable future). Your pull request just replaces english lines with turkish :)

HamzaYslmn:
I did localization without a language config file because it would have taken too much time to rewrite the application and create separate configs for each language.

I don't have time to try to do this for each version :D

So, replacing the English texts with their Turkish versions was the quickest solution

PR #195, opened 2023-04-05
Add a check for whether the dGPU is in use. Stop eco mode but allow a…

TheToby1:
… force disable.

### DIFF
--- a/app/HardwareMonitor.cs
+++ b/app/HardwareMonitor.cs
@@ -48,6 +48,20 @@ public static class HardwareMonitor
             return " Fan: " + Math.Min(Math.Round((float)fan/fanMax*100), 100).ToString() + "%"; // relatively to 6000 rpm
     }
 
+    private static int GetGpuUse()
+    {
+        try
+        {
+            int? gpuUse = GpuTemperatureProvider?.GetGpuUse();
+            if (gpuUse is not null) return (int)gpuUse;
+        }
+        catch (Exception ex)
+        {
+        }
+
+        return 0;
+    }
+
     public static void ReadSensors()
     {
         batteryDischarge = -1;
@@ -84,20 +98,8 @@ public static class HardwareMonitor
         if (gpuTemp is null || gpuTemp < 0)
             gpuTemp = Program.wmi.DeviceGet(ASUSWmi.Temp_GPU);
 
-        try
-        {
-            gpuUse = GpuTemperatureProvider?.GetGpuUse();
-        } catch (Exception ex)
-        {
-            gpuUse = -1;
-            Debug.WriteLine(ex.ToString());
-        }
-
-        if (gpuUse is not null && gpuUse >= 0)
-        {
-            gpuUsage.Add((int)gpuUse);
-            if (gpuUsage.Count > 3)  gpuUsage.RemoveAt(0);
-        }
+        gpuUsage.Add(GetGpuUse());
+        if (gpuUsage.Count > 3)  gpuUsage.RemoveAt(0);
 
         try
         {
@@ -114,8 +116,10 @@ public static class HardwareMonitor
 
     public static bool IsUsedGPU(int threshold = 50)
     {
-        if (gpuUsage.Count < 2) return false;
-        return (gpuUsage.Average() > threshold);
+        if (GetGpuUse() > threshold) 
+            return true;
+        else 
+            return (gpuUsage.Average() > threshold);
     }
 
     public static void RecreateGpuTemperatureProviderWithDelay() {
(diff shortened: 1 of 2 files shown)


TheToby1:
This isn't polished or feature complete, If you like the idea and want to make changes or have suggestions I'd happily accept/implement them.

Aurel100:
I had the same issue. This would be a great addition!

seerge:
Hello, thanks for the commit.

I have checked what AMD usage sensor reports and in normal use it seem to fluctuate a lot between literally 0% up to 20% in my case even on regular desktop use (cause windows itself uses dGPU as well). 

It makes preventing dGPU switching based on GPU usage % quite questionable (even if I will set a limit of usage not to 0 but to something higher like 30%).

Why don't you use Standard mode instead when you need dGPU and work with "unpredictable" power? :)

Aurel100:
i am doing the standard mode when on "unpredicable" power. 
I think the desired behavior is :
when unpredicable power happens => keep using GPU until it is closed by user

seerge:
> I think the desired behavior is :
> when unpredicable power happens => keep using GPU until it is closed by user

There is no definition for "unpredictable power" in windows :) if power is off, it's off. Current app behavior is transparent. Optimized mode - does switch, Standard doesn't.

TheToby1:
Interesting that your dGPU usage fluctuates, mine sits at 0 even with light tasks like browser usage, unless I launch something which uses the dGPU like YouTube, Unity or a game. The percentage could be a config along with whether this behaviour happens at all? Or I can check for a better way to check for running processes.

The problem is that I don't have "unpredictable" power, my apartment just had a fluke power cut. This time it was a game that crashed and I lost some progress, but it could just have easily been work.

TheToby1:
I found this https://github.com/GPUOpen-LibrariesAndSDKs/display-library/tree/master/Sample/PXRunningApps/PXRunningApps

If you like the idea of this feature I can look at basing it on running apps as above. I can also make it a popup alert as you mentioned rather than the current implementation.

seerge:
I tried to combine GPU usage check that you added, with alert (you can check latest main branch).

I have also added a logic for checking if GPU is in use

Instant use has to be > threshold (50%) OR average from last 3 checks has to be > threshold  

But i'm still not sure if it's needed tho :) you can check how it works in a build below

[GHelper.zip](https://github.com/seerge/g-helper/files/11161752/GHelper.zip)

seerge:
There are few unresolved problems still tho 
1. Nvidia GPU usage
2. Average GPU usage is updated only when RefreshSensors is called, and it's called ONLY when main app window is visible (on purpose, to consume less resources when app is in tray)

TheToby1:
That looks good. I'll test it later today.

1. I have a 2020 model I can look into getting the Nvidia GPU usage. 
2. If we could get running process info and it was useful, do you think this should remove the need for averages? I can look at setting this up for AMD and Nvidia.

seerge:
I have just added nvidia gpu usage (presumably), can you test it from main branch ?

seerge:
2020 models don't support disabling dGPU at all (on hardware level), it's even in app FAQ.

As for spanish - you probably caught main branch in the middle of some testing commits, now it should use system language.

As for GPU utilization - i'm not a big fan of lowering  limit, and it can create too much false alerts then. For example in my use case, i have external monitor connected (to dGPU) and even watching youtube in chrome easely fluctuates usage between 5% to 25% (rarely)

Logic of 50% - is to prevent app from turning dGPU while gaming and sudden power off (as you explained initially). In those situations you will almost certainly have 50% use

TheToby1:
"2020 models don't support disabling dGPU at all (on hardware level), it's even in app FAQ." - Ah my bad.

I understand the logic, I'm telling you that this fixed threshold only works for your usecase. In order to hit that threshold I had to launch Destiny 2 and not alt tab away, Running Unity as a dev, a small game or even Destiny but in the background would not push it over 50%.

"Separately using the release version downloaded from your assets (0.48) the warning never displays for me on the 2022 model." - Do you see this behaviour as well? I'd rather not have to build this tool locally in order to get it working.

seerge:
@TheToby1 i can propose following

Threshold will be lowered to 20, but it will wait for 1 more second and if it's still > 20 - then trigger alert

```
    public static bool IsUsedGPU(int threshold = 20)
    {
        if (GetGpuUse() > threshold)
        {
            Thread.Sleep(1000);
            return (GetGpuUse() > threshold);
        } else
            return false;
    }
```

TheToby1:
That would probably suit me better yes, but again I feel like fitting this threshold to every use case is never going to suit everyone.

seerge:
@TheToby1 well, 20 is low enough for you, right ? :)

TheToby1:
Haha, yes I think so. That release seems to be working!

Thanks for all the work by the way, I do appreciate it and I know I'm nit picking these things. If you have anything specific things you'd like tested or help with dev on I'd love to continue contributing.

PR #199, opened 2023-04-06
Update Fans.Designer.cs

lswlc33:


seerge:
Hello, that's cool. But currently app doesn't have  multilingual support :) your changes will simply replace all english text with chinese.

I may consider adding multilinguality later, but after UI will stabilize, and it will be clear what needs to be translated.

PR #215, opened 2023-04-10
Update debloat.bat

ArnavVarshney:
Add ASUS Cert Service to debloat. No online reference to its usefulness.


seerge:
Hello, current idea of debloat - is to disable only things installed with Asus System Control interface (as it's needed for g-helper to work properly, but installs a bit more than needed)

I may write a more wide script later that cleans up all leftower asus folders / etc, but that's for later:)

ArnavVarshney:
Fair enough, closing it then.

PR #237, opened 2023-04-13
Add fan levels auto-adjust

hungqbui:
G-Helper has been very useful since I made the switch from Armoury Crate. But tweaking with the fan levels have been annoying because they don't auto adjust. This is my first open source contribution ever so I'm sorry if my implementation is very elementary. Also, if there are any GitHub etiquettes that I'm not aware of please let me know.

### DIFF
--- a/app/Fans.cs
+++ b/app/Fans.cs
@@ -430,6 +430,13 @@ namespace GHelper
                     {
                         curPoint.XValue = dx;
                         curPoint.YValues[0] = dy;
+
+                        if (hit.Series is not null) {
+
+                            AdjustAllLevels(hit.PointIndex, dy, dx, hit.Series);
+
+                        }
+
                         tip = true;
                     }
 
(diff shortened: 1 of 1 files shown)


app/Fans.cs
```diff
@@ -450,6 +457,75 @@ private void ChartCPU_MouseMove(object? sender, MouseEventArgs e)
...
+            // Get the neighboring DataPoints of the hit point
+            DataPoint upperPoint = null;
+            DataPoint lowerPoint = null;
+
+            if (index > 0) 
+            {
+                lowerPoint = series.Points[index - 1];
+            }
+
+            if (index < series.Points.Count)
```
manciuszz, line 472: Probably going to break the graph tooltips because of out of bounds exception when interacting with the last node in the graph.

manciuszz:
I also see O(log n) potential over the current O(n) in-efficient code. However, that's just food for thought.

hungqbui:
Yeah I definitely think there are way better ways to optimize it. The code itself isn't very DRY either. I'll try to figure it out, probably going to take me a while though. Thanks for the suggestion!

manciuszz:
> Yeah I definitely think there are way better ways to optimize it. The code itself isn't very DRY either. I'll try to figure it out, probably going to take me a while though. Thanks for the suggestion!

Benchmark. Maybe play around with a Binary Search Tree. Benchmark again. I doubt there's going to be a significant difference considering we are dealing with a very low amount of graph nodes (points). However, if you never used any data structures before, here's your chance.

hungqbui:
I haven't really learned Data Structures, at least not formally. But I'll look into using BST. I can't find a better implementation of this for an array.

seerge:
@hungqbui hello, thanks for contribution, I will check it :)

hungqbui:
> @hungqbui hello, thanks for contribution, I will check it :)

You're welcome. It is definitely not the best implementation though, but I haven't found anything better at the moment.

PR #251, opened 2023-04-15
Prevent 'ImageAnimator' from creating an unnecessary thread

manciuszz:
Load .gif as an Image to prevent 'ImageAnimator' from creating an animation thread that basically is never used.

**Ideally, it would be best to get rid of that .gif entirely.**


seerge:
Well, it just stops gif from playing ? :)

manciuszz:
When you load 'Fans + Power' window, it would make 'ImageAnimator' create a worker 'animationThread' for animations - even if that **.gif** is never going to be played. It would be running in the background forever, until the app is closed. Closing 'Fans + Power' window doesn't 'dispose' of that thread. Hence the only way is to kill that thread using reflection or just prevent it from starting altogether.

seerge:
Oke, thanks. Closing this PR cause GIF is removed all together.

PR #325, opened 2023-05-06
Updating GHelper through a batch file

flankytingz:
This batch file updates G-Helper according to the link passed as a parameter.
Note: This only updates the GHelper.exe
This batch file could be executed through the program while passing the update GitHub link to update the application.

Expected run would be:
`.\Update.bat https://github.com/seerge/g-helper/releases/download/v0.57/GHelper.zip` through powershell/cmd


flankytingz:
I have created two separate files instead so the different update channels can be accessed.

`updateLatest.bat` will update the GHelper to the latest release automatically and does not require any arguments.

`updateSpecific.bat` will update the GHelper to the version number passed as the argument.
For example, to update to v0.57 the execution will be:
`.\updateSpecific.bat v0.57`

Along with that, I have improved the error handling and logging so if it fails to extract or replace, the batch file ensures to clean up and exit.

ke1ne:
@flankytingz I would propose to merge them to the single file, something like:
- running `updateLatest.bat` gets the latest stable version be default
- running `updateLatest.bat --ver=0.57` gets the specific version set in the parameter

flankytingz:
I have made changes according to your recommendation @ke1ne

`.\update.bat` -> downloads latest version, v0.55
`.\update.bat --ver=0.57` -> downloads version v0.57

PR #375, opened 2023-05-13
Supports AniMeMatrix for Zephyrus M16 2023 (GU604)

IceStormNG:
Update to the AniMeMatrix code to also support the Zephyrus M16 2023 (GU604). 
I tested it on mine with some GIFs and the Clock. Both look correct not and are no longer scrambled.

### DIFF
--- a/app/AnimeMatrix/AnimeMatrixDevice.cs
+++ b/app/AnimeMatrix/AnimeMatrixDevice.cs
@@ -61,7 +61,8 @@ namespace Starlight.AnimeMatrix
     public enum AnimeType
     {
         GA401,
-        GA402
+        GA402,
+        GU604
     }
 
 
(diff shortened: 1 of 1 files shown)


seerge:
Thanks for PR, are you sure about that "frameShiftX" btw ? it could be there are some "dead" pixels in first X rows that just don't show but take space in the array

IceStormNG:
> are you sure about that "frameShiftX" btw ? it could be there are some "dead" pixels in first X rows that just don't show but take space in the array

Yes. I tested it a few times, and the display wraps the lines differently than the G14 AnimeMatrix versions. There is always a shift of 5 LEDs on the X axis. If you do not correct that, you can see that the picture/gif wraps around the edges and shows up on the opposite. As for the invisible pixels. I don't think so. I counted them actually by hand and the numbers in my code and the number on the Matrix do match when I control them. It just has this 5 LED shift.

I tried to compensate without introducing a new variable, but no luck. Either shifts the image or scrambles it completely. This way it works without any visual issues. And as the default is 0, it does not affect any other models.

seerge:
@IceStormNG ok, cool :) I just had a loooot of pain adapting code to fit old GA401 without physical device just by someone's photos, untill figured out it was some dead pixels there (that's why Pitch(int y) exists)

PR #385, opened 2023-05-15
Adjustment of Windows Power Overlay is now optional

IceStormNG:
Right now, GHelper automatically adjusts the Windows Power Scheme Overlay (Best Energy Saving, Balanced, Best Performance) when you switch performance mode. 
I added a checkbox to make this change optional as it might not be desired. The default (for new installations) is "Enabled" which is identical to the current behavior. If you uncheck the box, GHelper does not modify the power scheme overlay anymore.

This only affects the change of power overlay. Other feature, like change of boost behavior, is unaffected by this.

This GHelper setting is not per performance mode but global.

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -69,6 +69,7 @@ public class AppConfig
     {
         config = new Dictionary<string, object>();
         config["performance_mode"] = 0;
+        config["auto_apply_power_plan"] = 1;
         string jsonString = JsonSerializer.Serialize(config);
         File.WriteAllText(configFile, jsonString);
     }
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -74,6 +74,7 @@ namespace GHelper
             checkSleepLid = new CheckBox();
             checkShutdownLid = new CheckBox();
             groupOther = new GroupBox();
+            checkApplyWindowsPowerMode = new CheckBox();
             checkKeyboardAuto = new CheckBox();
             checkUSBC = new CheckBox();
             checkNoOverdrive = new CheckBox();
--- a/app/Extra.cs
+++ b/app/Extra.cs
@@ -73,6 +73,7 @@ namespace GHelper
             checkNoOverdrive.Text = Properties.Strings.DisableOverdrive;
             checkTopmost.Text = Properties.Strings.WindowTop;
             checkUSBC.Text = Properties.Strings.OptimizedUSBC;
+            checkApplyWindowsPowerMode.Text = Properties.Strings.ApplyWindowsPowerPlan;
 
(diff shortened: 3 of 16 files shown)


seerge:
@IceStormNG hello, i don't mind to merge this , but what is practical approach of one more extra setting? Why would you want to unpair windows "Power mode" from Bios power mode ? (i think it's better to call it like that and not power plan, cause it will raise lots of questions) 

Even armoury forcely sets own power plans with each mode.

IceStormNG:
The reason is, that, at least on Intel laptops, the Windows Power mode has a great impact on CPU performance. "Best power saving" which is activated by default on Silent, really caps the CPU performance and disables turbo. Some people might want this, but I personally do not want that as it makes a lot of CPU heavy apps really sluggish. I prefer it to be set on balanced (windows mode) all the time and set the machine to silent (BIOS Mode) when I want the fans to be quiet and use balanced (BIOS power mode) when I don't need absolute silence. 
This is actually something that already bothered me about AC as it also coupled the windows power modes with the BIOS power modes and regularly reset them on several occasions. 
That's why I wanted to have an option to disable that. So people that like this feature can keep it, and people like me, that do not like it, can disable it.
We can of course rename the function if you have a better name that is less confusing.

seerge:
@IceStormNG oke, i agree with argumentation. But I need some time to think on best UI solution for that.

IceStormNG:
> @IceStormNG oke, i agree with argumentation. But I need some time to think on best UI solution for that.

I also wasn't really sure where to put it and how to name it. Let me know when you found a good way to integrate it in the UI and I will update the code.

seerge:
@IceStormNG on Fans screen everything is linked to currently selected mode , so it should be a dropdown of all 3 power modes with pre-selected one that app currently uses.

OR

I would add a checkbox like you did, but somewhere deep and hidden (under Extra as i do with some strange options people ask sometimes) :)

IceStormNG:
For me and my use case, the checkbox is all I need. So I will move it to the extras section.

IceStormNG:
The commit was not complete yet. I have to fix some UI issues as my VS screwed up the UI scaling due to HiDPI

seerge:
yeah, i just merged - checked and realised :)

seerge:
@IceStormNG i have reverted it back to previous state.

Since you decided to finish this
- I would rename it to "Power Modes"
- Remove `` config["auto_apply_power_plan"] = 1;`` from init config completely (so default value is nothing, as it will be for all people who will just update app)
- Adjust logic in settings.cs to ``if (Program.config.getConfigPerf("auto_apply_power_plan") != 0)`` (to accomodate case when auto_apply_power_plan is not set) :)

IceStormNG:
Alright. I will make some changes and re-create the pull request

PR #388, opened 2023-05-15
Feature: Auto update GHelper

flankytingz:
Added UI for auto update option under Extra.
Added update.bat for updating GHelper.exe


seerge:
I actually added built-in auto updater there (without bat file) :)

flankytingz:
I just saw that. =D finally.

PR #390, opened 2023-05-16
Auto update now deletes the zip file after installation.

flankytingz:


seerge:
@flankytingz thanks for PR :) 

I have added your change, but manually cause i think it's better to remove zip before running exe. As currently new exe on start will kill old one (which may interrupt script before) finishing.

PR #392, opened 2023-05-16
Windows power plan changes optional

IceStormNG:
Fixed the incomplete PR [https://github.com/seerge/g-helper/pull/385 ](https://github.com/seerge/g-helper/pull/385 )

Right now, GHelper automatically adjusts the Windows Power Scheme Overlay (Best Energy Saving, Balanced, Best Performance) when you switch performance mode.
I added a checkbox to make this change optional as it might not be desired. If you uncheck the box, GHelper does not modify the power scheme overlay anymore.

This only affects the change of power overlay. Other features, like change of boost behavior, are unaffected by this.

This setting is not per performance mode but global.

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -69,7 +69,6 @@ public class AppConfig
     {
         config = new Dictionary<string, object>();
         config["performance_mode"] = 0;
-        config["auto_apply_power_plan"] = 1;
         string jsonString = JsonSerializer.Serialize(config);
         File.WriteAllText(configFile, jsonString);
     }
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -69,6 +69,7 @@ public class AppConfig
     {
         config = new Dictionary<string, object>();
         config["performance_mode"] = 0;
+        config["auto_apply_power_plan"] = 1;
         string jsonString = JsonSerializer.Serialize(config);
         File.WriteAllText(configFile, jsonString);
     }
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -74,7 +74,7 @@ namespace GHelper
             checkSleepLid = new CheckBox();
             checkShutdownLid = new CheckBox();
             groupOther = new GroupBox();
-            checkApplyWindowsPowerMode = new CheckBox();
+            checkAutoApplyWindowsPowerMode = new CheckBox();
             checkKeyboardAuto = new CheckBox();
             checkUSBC = new CheckBox();
             checkNoOverdrive = new CheckBox();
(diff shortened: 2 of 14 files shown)


seerge:
@IceStormNG is this one "final" ? :)

IceStormNG:
> @IceStormNG is this one "final" ? :)

Yes. This one is final and should work. The checkbox is now in the "Extra" Window. Translations will need to be update. I put the English text in them as placeholder.

seerge:
@IceStormNG hello,

Topic for discussion : few times i have spotted people on reddit complaining / not understanding how power mode changes happen (and what's that in a frist place). Also some mention that in AC both Turbo & Performance modes use Max power setting (de facto),  while in G-Helper Balanced will use Balanced power mode.

Personally for me, current behavior is perfect, but I was thinking about making whole power mode changes more transparent and editable.

I wanted to add a dropdown to Fans+Power -> CPU section (above Boost) mentioning associated power mode .

What do you think?

IceStormNG:
> Topic for discussion : few times i have spotted people on reddit complaining / not understanding how power mode changes

How unexpected. I read so much weird things about this I already stopped trying to correct them. People also don't understand how modern standby work and how they can configure it to actually work (tbf: the OEMs do not really care about that either).

I mean. You can of course add that if you want to. Or: Just leave it as it is (power users can change it in the config.json) or they can use the toggle if they don't want GHelper to interfere with it.

Of course, we can add the whole power plan there. but that would slowly bloat up GHelper quite a lot.

seerge:
Examples

https://www.reddit.com/r/Asustuf/comments/15kp3ad/dont_use_g_helper_if_you_have_an_asus_laptop/
https://www.reddit.com/r/Asustuf/comments/15meyxj/ghelper_problem/

IceStormNG:
Maybe list it in the Readme. Sure, most people will not read, but that is not your fault. The defaults on GHelper are actually better to me than the ones of AC. 
Although I personally prefer "balanced" all the time. Best performance actually ignores a lot of power plan settings and sets them to the highest performance one (except for core parking. Coreparking cannot be disabled with max performance, no matter what you set in power plan, it still parks cores).

seerge:
@IceStormNG people don't read, and it's already listed https://github.com/seerge/g-helper#rocket-performance-modes. People don't read. Thanks to app being quite simple, it's being used by lots of people now, often not super PC savvy. And average Joe barely manages to unzip the folder properly

Problem that such topics (as well as others complaining about "not working charge limit" appearing like every day, scare new users away. As no one will read or dig into details. They just get impression that "app is bugged" (which is not true in a first place)

IceStormNG:
I know they don't read but that is their fault (I have this issue with some other content that I release. They do not read and then complain and ask questions that is actually just 1 click and 30s of reading away from them). 

Look. You cannot really make your app idiot proof. The universe will just answer with "better" idiots.

If you feel the need to implement it either way, then sure. Put it next to the boost setting or so.

PR #527, opened 2023-06-03
Update Spanish translation

marcelomijas:
- Translation for the new Fn+F without Fn feature.
- Translation for the Fn-Lock toggle.
- Update backlight timeout translation.


seerge:
Thanks!

PR #598, opened 2023-06-11
Detect screen frequency if possible for OD buttons

Gallardo994:
Currently it says 120HZ no matter what display one may have. We need a stable way of detecting true screen refresh rate.

One problem remains is that if the laptop has Mux Switch, and it has been set to dGPU-only mode, we cannot detect laptop screen properly as it's detected as external.

I have observed that the screen name is exactly the same as the internal screen name, but with the EXTERNAL tag at the end of the name string. Can we somehow use external-but-internal screen if primary screen is not detected perhaps?

### DIFF
--- a/app/NativeMethods.cs
+++ b/app/NativeMethods.cs
@@ -653,7 +653,7 @@ public class NativeMethods
 
             var screens = Screen.AllScreens;
 
-            if (screens.Length != count) return Screen.PrimaryScreen.DeviceName;
+            if (screens.Length != count) return null;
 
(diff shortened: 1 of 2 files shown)


Gallardo994:
Instead of falling back to "120 HZ + OD" text if Mux has been switched to dGPU, I have implemented falling back to primary screen refresh rate information. Works properly on my machine both when dGPU is enforced and when it's in Optimus mode. Might need extra testing on other devices, tested on G733PY.

PR #702, opened 2023-06-30
Updated German translation

IceStormNG:
I updated the German translation to include the recent additions and features.


seerge:
@IceStormNG great, thanks 🙏

PR #813, opened 2023-07-18
Added hotkeys for brightness control.

jadonclegg:
Added hotkeys for using ctrl / shift + vol up / down to control screen / keyboard brightness.

IE - CTRL + Vol Up = screen brightness up, CTRL + Vol Down = screen brightness down. SHIFT controls keyboard brightness.

Trying to adjust the keyboard brightness or screen brightness is painful on the laptop, finding the right keys in the dark / without having to look is difficult. Using CTRL  or SHIFT+ vol up / vol down is much easier to do than FN+F? combo.


seerge:
@jadonclegg hello, that's a reasonable idea, let's merge it. Thanks. 
(Even tho for me it's almost equally difficult to find M1/M2 as it would be finding F2/F3 for backlight control for example) :)

TheBestPessimist:
This feature sounds amazing!

PR #817, opened 2023-07-19
Automatic Clamshell Mode Toggle

IceStormNG:
This is something less specific to the Zepyhrus line but something that I personally want and others might also like without having to resort to other tools (if they even exist).

If you've used a MacBook, you likely know about Clamshell Mode. When charger, external display, mouse and keyboard are connected and you close the lid, nothing happens. Means: The machine will not go to sleep and can be used as a desktop replacement.

Windows does not support it like that. Windows has a setting for what to do when closing the lid. The setting does not care whether you have a display connected or not, though. You would have to change that setting manually every time, as sometimes you want it to go to sleep when you close the lid (with nothing besides the charger connected) and sometimes you want to use clamshell mode.

I added an optional toggle to replicate clamshell mode. It does the following:
When charger and at least 1 external display is connected, it toggles the windows power setting for the lid action on charger to "Do nothing". If the display is disconnected, that setting is reverted to "Sleep". I omitted the external keyboard and mouse detection for simplicity sake and because on MacBooks it sometimes caused issues if macOS hasn't detected your keyboard as such correctly. It also does not touch the settings for "on battery".

The toggle is in the "Extras" section.

I tested it on my GU604VZ (M16 2023), but it should work the same on all other models and more or less any Windows laptop.

### DIFF
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -101,6 +101,7 @@ namespace GHelper
             pictureSettings = new PictureBox();
             labelSettings = new Label();
             panelSettings = new Panel();
+            checkAutoToggleClamshellMode = new CheckBox();
             checkAutoApplyWindowsPowerMode = new CheckBox();
             checkTopmost = new CheckBox();
             checkNoOverdrive = new CheckBox();
@@ -1003,6 +1004,7 @@ namespace GHelper
             // 
             panelSettings.AutoSize = true;
             panelSettings.AutoSizeMode = AutoSizeMode.GrowAndShrink;
+            panelSettings.Controls.Add(checkAutoToggleClamshellMode);
             panelSettings.Controls.Add(checkAutoApplyWindowsPowerMode);
             panelSettings.Controls.Add(checkTopmost);
             panelSettings.Controls.Add(checkNoOverdrive);
@@ -1016,6 +1018,20 @@ namespace GHelper
             panelSettings.Padding = new Padding(20, 5, 10, 5);
             panelSettings.Size = new Size(983, 304);
             panelSettings.TabIndex = 46;
+            // 
+            // checkAutoToggleClamshellMode
+            // 
+            checkAutoToggleClamshellMode.AutoSize = true;
+            checkAutoToggleClamshellMode.Dock = DockStyle.Top;
+            checkAutoToggleClamshellMode.Location = new Point(10, 163);
+            checkAutoToggleClamshellMode.Margin = new Padding(2);
+            checkAutoToggleClamshellMode.Name = "checkAutoToggleClamshellMode";
+            checkAutoToggleClamshellMode.Padding = new Padding(2);
+            checkAutoToggleClamshellMode.Size = new Size(481, 23);
+            checkAutoToggleClamshellMode.TabIndex = 58;
+            checkAutoToggleClamshellMode.Text = "Auto Toggle Clamshell Mode";
+            checkAutoToggleClamshellMode.UseVisualStyleBackColor = true;
+            checkAutoToggleClamshellMode.CheckedChanged += checkAutoToggleClamshellMode_CheckedChanged;
             // 
             // checkAutoApplyWindowsPowerMode
             // 
@@ -1246,5 +1262,6 @@ namespace GHelper
         private PictureBox pictureService;
         private Slider sliderBrightness;
         private PictureBox pictureLog;
+        private CheckBox checkAutoToggleClamshellMode;
     }
 }
--- a/app/Extra.cs
+++ b/app/Extra.cs
@@ -106,6 +106,7 @@ namespace GHelper
             checkUSBC.Text = Properties.Strings.OptimizedUSBC;
             checkAutoApplyWindowsPowerMode.Text = Properties.Strings.ApplyWindowsPowerPlan;
             checkFnLock.Text = Properties.Strings.FnLock;
+            checkAutoToggleClamshellMode.Text = Properties.Strings.ToggleClamshellMode;
 
             labelBacklightKeyboard.Text = Properties.Strings.Keyboard;
             labelBacklightBar.Text = Properties.Strings.Lightbar;
@@ -209,6 +210,9 @@ namespace GHelper
                     labelBacklightKeyboard.Visible = false;
                 }
             }
+
+
+            checkAutoToggleClamshellMode.Checked = AppConfig.Is("toggle_clamshell_mode");
 
             checkTopmost.Checked = AppConfig.Is("topmost");
             checkTopmost.CheckedChanged += CheckTopmost_CheckedChanged; ;
@@ -465,6 +469,13 @@ namespace GHelper
         private void checkAutoApplyWindowsPowerMode_CheckedChanged(object? sender, EventArgs e)
         {
             AppConfig.Set("auto_apply_power_plan", checkAutoApplyWindowsPowerMode.Checked ? 1 : 0);
+        }
+
+        private void checkAutoToggleClamshellMode_CheckedChanged(object sender, EventArgs e)
+        {
+            AppConfig.Set("toggle_clamshell_mode", checkAutoToggleClamshellMode.Checked ? 1 : 0);
+            ClamshellModeControl ctrl = new ClamshellModeControl();
+            ctrl.ToggleLidAction();
         }
     }
 }
--- a/app/Helpers/ClamshellModeControl.cs
+++ b/app/Helpers/ClamshellModeControl.cs
@@ -1,0 +1,76 @@
+﻿using GHelper.Display;
+using GHelper.Mode;
+using Microsoft.Win32;
+
+namespace GHelper.Helpers
+{
+    internal class ClamshellModeControl
+    {
+        public bool IsExternalDisplayConnected()
+        {
+            var devices = ScreenInterrogatory.GetAllDevices().ToArray();
+
+            foreach (var device in devices)
+            {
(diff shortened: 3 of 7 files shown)


seerge:
@IceStormNG oke, we can try that as well :) I'm merging then.

seerge:
@IceStormNG seem to work :) the only thing control panel in windows itself seem to be bugged and keep showing "sleep" as action (i assume it doesn't re-read power plan settings)

Screenshot 2023-07-19 122610

IceStormNG:
> @IceStormNG seem to work :) the only thing control panel in windows itself seem to be bugged and keep showing "sleep" as action (i assume it doesn't re-read power plan settings), but laptop doesn't go to sleep when lid is closed + external monitor is connected

It does show the changes for me, but you have to open the setting again. So the control panel does not live update. If you go one view back and open this view again, it should show the changes. If not, then it hasn't detected your display as external. There is some logging I added, but maybe I should add an indicator to GHelper to tell the user that it is engaged or not.

seerge:
@IceStormNG my control panel seem to bug in general, it doesn't even show a power plan by default , so i guess that's problems specific to my windows only :) 

Screenshot 2023-07-19 124312

IceStormNG:
Oh. That could be true. The setting seems to be related to the power plan. I change it for the active power plan. So if that is bugged then, of course that settings will also bug out.

seerge:
Ok, i managed to fix my own issues by creating a temp plan, going there and back, and deleting it. Now control panel seem to listen to "Balanced" power plan settings and reflect them as it should 

By the way, since you touched topic of adjusting power plan. I was thinking about adding some kind of "optimize power plan" method in app that would set min processor state to 5% for both AC/DC, and probably some other minor things that people seem to do anyway. I'm just not sure if it won't be too invasive ... :) 

Screenshot 2023-07-19 124819

IceStormNG:
I was also thinking about that. There are also some settings in there that affect sleep stability with modern standby. Something that people have issues with, but I fixed on my machine. So a simple and accessibly solution for typical users would probably be not a bad idea.
Same thing with "Processor efficiency settings" or "Core Parking".  Windows hides most settings by default.

AudreyAP:
This is so cool. Thank you!

If one unplugs the display / power, will the laptop immediately sleep? Worried about having it maybe run full blast while in a backpack etc.

IceStormNG:
> This is so cool. Thank you!
> 
> If one unplugs the display / power, will the laptop immediately sleep? Worried about having it maybe run full blast while in a backpack etc.

I just tested that. Sadly, it does not. But I can add an option so that it does that when clamshell mode gets disabled and the lid is closed.

seerge:
@AudreyAP i see that it actually disables clamshell mode when power / screen is disconnected while laptop is closed, but windows indeed don't go to sleep as it doesn't get any trigger for that (as lid is already closed ...) :) 

@IceStormNG it's irrelevant, but i see that some other guy edited some German translations - can you check if that's relevant (feel free to click Undo if it's not ) https://crowdin.com/project/g-helper/activity-stream (i just don't see how to restrict project, and approve translators manually ...)

IceStormNG:
I guess this will not work. When I disconnect the display, the system sends a "Lid opened" event. So windows thinks the Lid is open when disconnecting the display while in clamshell mode.  The Lid open event arrives just before the display change event.

There is no way to actually query the physical lid state. This is all well hidden in the ACPI and SMBIOS and Windows keeps that for itself.

AudreyAP:
@IceStormNG So you mean that there is no way to tell the difference between unplugging a display and opening the actual display? That's annoying...

AudreyAP:
https://techcommunity.microsoft.com/t5/windows-powershell/get-lid-open-or-close-state-using-powershell/m-p/2398308

Found this... Unsure if useful or if you've seen it but this seems to abstract the lid sensor away to the display itself?

IceStormNG:
The problem with that is, that the internal display becomes online as soon as the external display gets disconnected. So the device turns on the internal display and sends a lid open signal.
There is no way to distinguish that from an actual case where you open the lid.

seerge:
@IceStormNG btw, i have found another small (and fixable) issue - when you have nvidia display mode in Nvidia GPU only it would create a virtual display in system for that, that also identifies itself as "external" and it confuses your clamshell detector :)

But it's possible to teach app to ignore that case, as it already saves correct name for "real" internal display in config (as soon as it spots it) so you can add an extra check if name of detected external display is NOT the one from config (as you want to get real external displays only)

https://github.com/seerge/g-helper/blob/e7a4537098873c3509ac364b5ee704d4e44e5052/app/Display/ScreenNative.cs#L158C27-L158C27

IceStormNG:
@seerge  yeah, I just thought about that. I already sent a pull request for the fix.

@AudreyAP As for the auto-sleep when closed: Either, there is some hidden info in the ASUS ACPI we can use (Armory crate somehow knows that the lid is closed as it can show a warning about that), or you will have to open the lid and close it again after using clamshell mode. Putting it to sleep and removing the external display afterwards will likely wake the computer as it sends a "lid open" signal. Silly Microsoft....

IceStormNG:
The good news. I found a possible way to solve it anyways. Apparently, the laptop emits WMI event 192 from ASUS WMI Events when a display is connected or disconnected (at least mine does it. And this even only comes with external display connects/disconnects for me).

Only issue right now. Even though I called to put it to sleep, it hibernated for whatever reason.

seerge:
@IceStormNG how do you call it to go to sleep?

IceStormNG:
> @IceStormNG how do you call it to go to sleep?

I tried

```cs
 [DllImport("Powrprof.dll", CharSet = CharSet.Auto, ExactSpelling = true)]
public static extern bool SetSuspendState(bool hiberate, bool forceCritical, bool disableWakeEvent);

public static void InitiateSleep()
{
    SetSuspendState(false, true, true);
}
````

But it looks like this does not work with modern standby. I found another solution but that doesn't work until the lid is opened once after the display unplug.

**Comment by seerge** [comment]
@IceStormNG try emiting FN+F11 hotkey as I do for FN-lock

```
Program.acpi.DeviceSet(AsusACPI.UniversalControl, AsusACPI.KB_Sleep, "Sleep");
```

IceStormNG:
> Program.acpi.DeviceSet(AsusACPI.UniversalControl, AsusACPI.KB_Sleep, "Sleep");

Works, but has the same issue as the other command that I found. It only works when you open the lid, then it goes immediately to sleep. But not during the closed lid. The thing is, the integrated display is "off" until the lid opens again. I guess this causes the issue.

Only other solution I found that supposedly works: Disable hibernate, initiate sleep and re-enable hibernate quickly again. Super goof, if you ask me.

seerge:
@IceStormNG i agree, I would suggest to keep things as is then.

IceStormNG:
> @IceStormNG i agree, I would suggest to keep things as is then.

It does work though if you unplug the charger. Then it goes to sleep immediately. Somehow I have to make Windows re-evaluate the power state.

seerge:
> if nothing else prevents it 
:D

AudreyAP:
@IceStormNG yay! The "unplug and backpack" problem is the only one I saw, so seems good :)

IceStormNG:
Well, I found a solution anyways, but that might need some tweaking. I call "DisplaySwitch.exe" (the tool that opens when you press WIN+P). It accepts command line parameters. I force it to Internal only, this re-enables the internal screen, and then it processes the sleep command and goes to sleep immediately.
Only issue is, I have to call this on WMI event before Windows even realized that the external screen is gone. I need to do additional checks to make sure that all external screens are gone (probably via WMI). Otherwise unplugging one external screen would force it to "internal only" and sleep. Which is not what you would want.

IceStormNG:
Ok. I give up. We will leave it like it is. The code I have works fine but gets confused once you add more than one monitor and disconnect one or add another one while in clamshell. Not worth the trouble. I cannot write against a broken Windows API.

Therefore, it might be good to add a help note to the FAQ that you have to pull the plug to make it sleep when you don't want to open the lid anymore. That works, I tested it a few times.

The problem seems to be that, the internal display is disconnected until you either remove the charger or open the lid. I tried to do that with displayswitch.exe, which works, but has the downside that I have to call it before the display change event. Once the last display is gone and only the disabled internal screen is left, I cannot do anything anymore. Whatever I call has no effect, as if Ghelper is paused due to having no output display. Could be a behavior from windows. When the change event happens for the display, it is already too late, but for the WMI event, it is too early as the display is still registered in WMI. The ASUS WMI event also doesn't say anything more than "Something changed on the displays". Nothing more.

seerge:
@IceStormNG one more thought about mode.

Currently it would subscribe to screen-change event and enable sleep on lid close even if checkbox is unchecked.

I would completely disable whole logic if checkbox is OFF (i.e. wouldn't touch power plan / detect screens / etc). Cause may be someone has tricky settings for their lid, and app would override them w/o any consent. 

What do you think?

IceStormNG:
> I would completely disable whole logic if checkbox is OFF (i.e. wouldn't touch power plan / detect screens / etc). Cause may be someone has tricky settings for their lid, and app would override them w/o any consent.

Yes. You're right. Currently it expects sleep on lid close as the default and sets that regardless. I will change that so the whole clamshell logic does not engage at all when the setting is disabled

seerge:
@IceStormNG cool. With the only remark. If checkbox was already ON and being unchecked - I would still set Sleep ON for lid-close 

This is for people who would check it w/o understanding what it does, and then uncheck back. So they are not being left with sleep off out of a sudden :)

PR #818, opened 2023-07-19
Fixed: Keyboard color cannot be set to white anymore

IceStormNG:
I just found out that with the latest pre-release (0.99.0.0), I cannot set my keyboard backlight color to white anymore. It always reverts to red. Every other color works fine.
After some debugging, I found out that white is encoded as "-1" by ToARGB().
This code was added by a recent commit, but it has the side effect of preventing white color. I'm not sure what this check is about, but -1 is not an invalid value, but pure white color (255/255/255).

### DIFF
--- a/app/AsusUSB.cs
+++ b/app/AsusUSB.cs
@@ -175,14 +175,12 @@ namespace GHelper
 
         public static void SetColor(int colorCode)
         {
-            if (colorCode == -1) Color1 = Color.Red;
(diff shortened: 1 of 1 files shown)


seerge:
@IceStormNG thanks for spotting :) it was connected with setting keyboard mode every time app runs / wakes up (as a fix for someone).

seerge:
Re-uploaded build with fix over existing https://github.com/seerge/g-helper/releases/tag/v0.99

PR #835, opened 2023-07-20
Added battery health reading to below the battery charge limiter bar.

IceStormNG:
I guess this is not that important, but it would be nice to know how much the battery is degraded. As we already read battery information and show charge/discharge, we can also read and compute battery health.

image

_(Yeah, my battery is already quite screwed after just 4 months of use. No idea why)_

### DIFF
--- a/app/HardwareControl.cs
+++ b/app/HardwareControl.cs
@@ -13,7 +13,7 @@ public static class HardwareControl
 
     public static float? cpuTemp = -1;
     public static decimal? batteryRate = 0;
-    public static decimal batteryWear = -1;
+    public static decimal batteryHealth = -1;
     public static decimal? designCapacity;
     public static decimal? fullCapacity;
     public static int? gpuTemp = null;
@@ -145,8 +145,13 @@ public static class HardwareControl
         }
     }
 
+    public static void RefreshBatteryHealth()
+    {
+        batteryHealth = GetBatteryHealth() * 100;
+    }
 
-    public static decimal GetBatteryWear()
+
+    public static decimal GetBatteryHealth()
     {
         if (designCapacity is null)
         {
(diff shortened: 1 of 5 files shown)


seerge:
@IceStormNG hello, cool, but i think it's too much reading that values literally every time sensors refresh (i.e. 2 seconds). 

May be read it once (on app start) and show as a tooltip (or as label, but make label visible only  when you mouseover battery section) ?

IceStormNG:
> @IceStormNG hello, cool, but i think it's too much reading that values literally every time sensors refresh (i.e. 2 seconds).
> 
> May be read it once (on app start) and show as a tooltip (or as label, but make label visible only when you mouseover battery section) ?

The reading itself shouldn't be an issue as it is through WMI and low overhead. We can make it a tooltip if you think it wastes space, though.

seerge:
@IceStormNG well, each extra reading is still resources. Battery wear doesn't change in realtime (actually it can change only on full discharge / charge cycle) :)

And for tooltip - i just want to keep main window as clean as possible (as it's already quite cluttered tbh) :)

IceStormNG:
Alright. Let me change that...

IceStormNG:
@seerge  what about this.
I use the discharge/charge rate label to show the health when no charging happens and when it charges/discharges you have to hover your mouse over it to see the health. Otherwise it shows the charge/discharge rate as usual.

seerge:
@IceStormNG you can use the label you have already added to keep things simple, just make it visible on mousemove over battery panel, and hide otherwise :) (like "fake" tooltips for GPU / Screen sections work)

IceStormNG:
This is normal:
image


And this is when hovered:
image

It only refreshes when it is shown and only when at least 15 minutes passed since the last refresh.

seerge:
@IceStormNG good! danke :)

seerge:
@IceStormNG i've thought about it a bit, and looks like your idea with re-using original discharge label can be better (or cleaner)

What do you think about this build ?
[GHelper.zip](https://github.com/seerge/g-helper/files/12111779/GHelper.zip)

IceStormNG:
> What do you think about this build ? [GHelper.zip](https://github.com/seerge/g-helper/files/12111779/GHelper.zip)

It works pretty weird. In the sense that it is unpredictable when the lable appears and when not. Actually, I kinda like the tooltip approach now that I see the label-reuse solution  :D

seerge:
@IceStormNG i have made disappear immediate + added Battery left label also to trigger list
[GHelper.zip](https://github.com/seerge/g-helper/files/12112822/GHelper.zip)

IceStormNG:
> @IceStormNG i have made disappear immediate + added Battery left label also to trigger list [GHelper.zip](https://github.com/seerge/g-helper/files/12112822/GHelper.zip)

This one is much  better.

PR #842, opened 2023-07-21
Clamshell mode fixes

IceStormNG:
This fixes some things about clamshell mode:

- It does not touch the settings at all if the settings is not checked
- If unchecked it will always revert to the system default (Sleep)
- Some internal code cleanup and exposing some functions

### DIFF
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -1031,7 +1031,6 @@ namespace GHelper
             checkAutoToggleClamshellMode.TabIndex = 58;
             checkAutoToggleClamshellMode.Text = "Auto Toggle Clamshell Mode";
             checkAutoToggleClamshellMode.UseVisualStyleBackColor = true;
-            checkAutoToggleClamshellMode.CheckedChanged += checkAutoToggleClamshellMode_CheckedChanged;
             // 
             // checkAutoApplyWindowsPowerMode
             // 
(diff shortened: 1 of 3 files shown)


seerge:
@IceStormNG by the way, 

I was hiding some unsupported bindings / checkboxes under Exta for different models, and had an idea to hide clamshell mode checkbox as well, if there is no external display connected (my point here is not to overload average user with options that he has no idea about)

What do you think ?

IceStormNG:
> What do you think ?

I would leave it there. The reason for the box is so that it automatically enables clamshell when you connect a monitor. Having it only there with a monitor enabled could confuse the user that they think they have to toggle it there manually every time.

seerge:
@IceStormNG someone on reddit complains about clamshell
https://www.reddit.com/r/ZephyrusG14/comments/159l1l2/comment/jtjqqxq/?utm_source=reddit&utm_medium=web2x&context=3

my assumption is that he had **hibernate** action on lid close (instead of sleep)

Does it make sense to save current action from powerplan into a config param when Clamshell is checked (instead of assuming that action was **sleep**?)

IceStormNG:
@seerge if we store the previous setting, there is a risk that it stores "do nothing" because for example that ghelper was closed after it set clamshell mode and then restarted or it crashed.
We could however add this to the "power user settings" and add config option for those that do not want sleep (which is likely only a few people).

PR #855, opened 2023-07-22
Eco mode switching instability fix

joe-700:
Fix for system instability when switching to eco mode, this prevents the need to force GPU drivers to reset.

The current BIOS for new G14s does not boot correctly into eco GPU mode, this results in loss of Nvidia display modes i.e. advanced optimus, and loss of brightness controls. Currently a GPU driver restart is used to correct this state but this operation can either: hang, result in loss of brightness controls, and will often require a couple of resets to fully rectify. I have noticed system instability after doing this including one blue screen crash.

This fix will ensure that the GPU mode switch is set to normal before system shutdown and hibernation in order to guarantee booting into a safe state.

Current user selected GPU modes are honored by setting flags in the newly created xml file and restoring the GPU state after booting correctly into normal mode.

### DIFF
--- a/app/BootFromEco.xml
+++ b/app/BootFromEco.xml
@@ -1,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Session>
+  <BootFromEco>False</BootFromEco>
+  <BootFromOptimised>False</BootFromOptimised>
+</Session>
--- a/app/BootFromEco.xml
+++ b/app/BootFromEco.xml
@@ -1,5 +1,0 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
-<Session>
-  <BootFromEco>False</BootFromEco>
-  <BootFromOptimised>False</BootFromOptimised>
-</Session>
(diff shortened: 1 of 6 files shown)


joe-700:
I have a modified version that uses the app config to track the eco mode shutdown

joe-700:
As above only using the app config to track eco mode shutdown

seerge:
@joe-700 hello, i can't accept your PR as it will be forcing standard mode for all devices.

But I have already implemented pretty much same thing as you wanted 
https://github.com/seerge/g-helper/commit/20b99810cb983d166dbf426c514b63fea4f8ae72

You need to set ``"gpu_fix" : 1, `` in config.json.

And it will force standard mode on shutdown / and session ending events. And restore last GPU mode on startup (that implementation was already in the app)

seerge:
@joe-700 check this build , it should do the same : 
[GHelper.zip](https://github.com/seerge/g-helper/files/12138627/GHelper.zip)

joe-700:
That build works a treat.

I went back and tried the current version, via the download link from the main page. That one does not work, are you intending to roll this into the main branch?

seerge:
@joe-700 I just pushed it live as 0.101
https://github.com/seerge/g-helper/releases/tag/v0.101

PR #887, opened 2023-07-25
Asus mouse support (at least for 2 mice for now)

IceStormNG:
I decided to merge this already. It works and is tested for those two mice:

- ROG Chakram X (P708) 
- ROG Strix III Gladius III Aimpoint Wireless (P711)

The Strix Carry (P508) might be added later, once we figure out how it works.

Currently only those two mice are fully supported (everything except button assignment) as I do not want to add mice that are not tested. It could crash the app for people that are using them.
We can always add more mice later once we got people that want to help us with it, or they can add it themselves if they know how to (it's not hard, just a lot of work).
Also: It only supports one mouse of each kind at the same time. Though I doubt this is going to be a huge problem, if at all.

For Reference: A list of all mice that Armoury Crate supports (extracted from the driver INF). Might be handy later.
[ASUS_Mice.txt](https://github.com/seerge/g-helper/files/12163091/ASUS_Mice.txt)

Screenshot 2023-07-25 172316

The "peripherals" panel is fully gone when no mouse is detected that is supported by GHelper.

Screenshot 2023-07-25 172308

@seerge Due to your ongoing UI changes, there were a lot of merge conflicts with the Settings.Designer.cs. I hope I haven't broken your UI too much while trying to resolve them.


IceStormNG:
That was fast :D

seerge:
@IceStormNG hello :) Thanks a lot. That's indeed big and impressive chunk of functionality that you have added! :)

seerge:
@IceStormNG ye, I will do a post in Discussion a bit later then, where we can explain what info do we need from users, who want to add their mouse to the app.

And btw, what scaling do you use in visual studio ? i can imagine 150%? (as it's editor so terrible, that it messes up with sizing's) :) It's not a blame, i just noticed that pretty much every element that was touched - got resized like in 1.5% times ... (I use 200%)

IceStormNG:
Yes. I do indeed run at 150%. I restartet it at 100% but it looked the same for me. 
Yeah. The editor is pretty weird sometimes and always causes huge amounts of changes when I touch the UI. No matter whether I run at 100% or 150%.

And yeah. The discussion might be good. I somehow feel that there are 2 or 3 different protocols around because the Carry, that MD has doesn't even read out the battery properly. (It's a P5xx) while the gladius and chakram are both P7xx. Maybe that is a hint to what protocol to use. I see P3xx, P5xx and P7xx mice in the driver.

seerge:
@IceStormNG ye, that's the reason I didn't wanted even to touch peripherals (by myself), as it's probably a lot of very different implementations

IceStormNG:
Understandable. But I wanted it anyways. I mean, the project is opensource, so people can add their implementation easily if they found out how their mouse works. I made the main implementation as expandable and customizable as possible by separating functions into small pieces and making everything protected virtual.
Armoury Crate is sadly a big pile of crap right now and super slow. And booting a VM to change mouse settings feels a little.. over the top.

IceStormNG:
Just one note.. right now, the logger logs the whole USB transfer. This can fill the log quickly, but it helps a lot for debugging purpose and finding issues in the mouse protocol. Maye we should have a const that enables or disable packet logging. Just something that came to my mind right now....

seerge:
@IceStormNG well, log auto-cleans after 1000 lines , so it shouldn't be a problem

IceStormNG:
> @IceStormNG well, log auto-cleans after 1000 lines , so it shouldn't be a problem

This is good to know.. I will still add an app config option to disable or enable the packet logger, with the logger being disabled by default. It's only needed if the mouse protocol bugs out and for testing new mice. Although a software like this: https://freeusbanalyzer.com/ (which I used) to log all USB traffic is probably better to find all the magic numbers.

alexatthehub:
I've got a TUF M4 Wireless. I haven't touched programming or visual studio in years, but the individual mouse cs in the models folder seems _mostly_ understandable to me. Looks like the PID I need is 0x19F4. Apart from putting together a cs file for my mouse and updating the list in PeripheralsProvider.cs, is there much more I'd have to do to get things up and running?

_If I'm way out of my depth here I can just wait for Seerge to post the general discussion._

IceStormNG:
> I've got a TUF M4 Wireless. I haven't touched programming or visual studio in years, but the individual mouse cs in the models folder seems _mostly_ understandable to me. Looks like the PID I need is 0x19F4. Apart from putting together a cs file for my mouse and updating the list in PeripheralsProvider.cs, is there much more I'd have to do to get things up and running?
> 
> _If I'm way out of my depth here I can just wait for Seerge to post the general discussion._


Well, we will look into that later once the discussion is open. The ID is correct, but you also need to find out the mouse interface, and then implement the class based on capabilities. The main fun begins when the mouse does not adhere to the same protocol as the Chakram, which is the reference implementation. For example the Strix Carry that I currently try to implement diverts quite a lot from that.
Though the packet logger makes it possible to find the values you need. Otherwise, you need a USB sniffer software and see what Armoury crate sends when you change certain settings and then put 1 and 1 together.

seerge:
@IceStormNG I have made a new post in discussions (and also added a link to it from releases page)

https://github.com/seerge/g-helper/discussions/900

Feel free to add more info there (if i missed something) :)

naixx:
@IceStormNG I've tested with Chakram X - works great, thank you! Any hints how to rebind buttons without AC? Custom buttons/macros and notification were two most used “features” of AC. Right now I've bound with AC and deleted, hopefully settings won't be lost.

IceStormNG:
> Any hints how to rebind buttons without AC

Not yet. Although I decoded the button assignment, I haven't really understood what the keycodes mean and whether they're portable across different mice.
I also haven't really though out how to make a proper GUI to set those things.

PR #904, opened 2023-07-26
Asus mouse fixes

IceStormNG:
Some fixes for the mouse settings.

- Prevents crashes if the mouse disconnects when the window is open.
- DPI profile is now stored properly so it stays when the mouse is restarted.
- Mouse updates are handled in background to not lock up the UI when the mouse is slow to respond.

### DIFF
--- a/app/AsusMouseSettings.cs
+++ b/app/AsusMouseSettings.cs
@@ -51,8 +51,6 @@ namespace GHelper
             Shown += AsusMouseSettings_Shown;
             FormClosing += AsusMouseSettings_FormClosing;
 
-            mouse.Disconnect += Mouse_Disconnect;
-            mouse.BatteryUpdated += Mouse_BatteryUpdated;
             comboProfile.DropDownClosed += ComboProfile_DropDownClosed;
 
             sliderDPI.ValueChanged += SliderDPI_ValueChanged;
@@ -83,7 +81,8 @@ namespace GHelper
             comboBoxAutoPowerOff.DropDownClosed += ComboBoxAutoPowerOff_DropDownClosed;
 
             InitMouseCapabilities();
-            RefreshMouseData();
+            Logger.WriteLine(mouse.GetDisplayName() + " (GUI): Initialized capabilities. Synchronizing mouse data");
+            Task task = Task.Run((Action)RefreshMouseData);
         }
 
         private void AsusMouseSettings_FormClosing(object? sender, FormClosingEventArgs e)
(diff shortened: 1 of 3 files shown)


seerge:
@IceStormNG sorry to intrude in your contribution, but I have caught g-helper in taking some resources while having main window open. It's partially due to a Heatmap mode (cause it refreshes / sets color once per second).

But I also looked into Peripherals calls. And tried to remove then from RefreshSensrors and put in other places a bit. Can you please double check if it's working (cause i don't have a single asus mouse to test)

https://github.com/seerge/g-helper/commit/5830220a3d3d9fad34ae165584dc8116b44cba35

1. I moved UpdateSettingsView() to DetectAllAsusMice() (+ it's called on connect/disconnect as well, as i see). As I understand it doesn't make sense to refresh it once per 2 seconds otherwise? :)
2. I have moved PeripheralsProvider.RefreshBatteryForAllDevices(); to moment when app window is shown (cause otherwise it would check for empty array / try to loop there also once per 2 seconds and in 99% cases i assume people don't have any mice connected)

Is that ok ?

IceStormNG:
It currently deadlocks with the "refreshallbattery" call and blocks the settings view from opening if a mouse is connected on startup. I will check that tomorrow why it does that.

seerge:
@IceStormNG ops, sorry. Thanks for checking

seerge:
Btw, it's absolutely separate topic - but does same thing happen on your M16 ? https://www.reddit.com/r/ZephyrusM16/comments/159t5x6/comment/jtk4cod/?utm_source=reddit&utm_medium=web2x&context=3

IceStormNG:
> Btw, it's absolutely separate topic - but does same thing happen on your M16 ? https://www.reddit.com/r/ZephyrusM16/comments/159t5x6/comment/jtk4cod/?utm_source=reddit&utm_medium=web2x&context=3

This happened to me once. Though, I also had other issues with GPU switching. When Advanced optimus went to Nvidia Only mode, the whole screen froze, but the PC was still running. For me, it was the Nvidia Driver. It only worked fine with the ASUS Nvidia Driver not with the generic one.
I can try it later again whether switching to Ultimate works. But: The Nvidia drivers, especially for the 4000 series cards are still super bugged and slow.

seerge:
@IceStormNG well, what they discuss is not directly linked to Nvidia as I understand. They say that you need to set Standard (i.e. enable dGPU) prior to rebooting to Ultimate.

AC won't let you set ultimate from eco neither, so I just want to figure out if it makes sense to block ultimate on M16 if eco is on or not ...

IceStormNG:
Hm... I can test it later what happens. But either way. I would block it. Or: Set the GPU first to standard and then to ultimate inside Ghelper if you notice that it goes from eco to ultimate.

seerge:
@IceStormNG yes, forcing standard prior to ultimate is one of the options that i was thinking about :) But it seem to be needed only for M16 so far as G14, G16, 16 Strix, and lots of other models handle this transition fine as is. Furthermore - G14 remembers what was "starting" mode, and when you disable Ultimate - it would return back to Eco (and not Standard)

IceStormNG:
It could be a BIOS bug. The M16 2023 is bugged af anyways right now. As much as I like the machine, it looks like ASUS rushed it out without any time for polish and error testing.

PR #905, opened 2023-07-26
Clamshell power user settings

IceStormNG:
This allows users to set their lid default action when clamshell is not engaged.

Set `clamshell_default_lid_action` in config.json to any of those values.

0 = Do nothing
1 = Sleep (default)
2 = Hibernate
3 = Shutdown

### DIFF
--- a/app/Helpers/ClamshellModeControl.cs
+++ b/app/Helpers/ClamshellModeControl.cs
@@ -6,6 +6,13 @@ namespace GHelper.Helpers
 {
     internal class ClamshellModeControl
     {
+
+        public ClamshellModeControl()
+        {
+            //Save current setting if hibernate or shutdown to prevent reverting the user set option.
+            CheckAndSaveLidAction();
+        }
+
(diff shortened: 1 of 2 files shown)


seerge:
Does it make sense to manually set this variable on check of "auto clamshell" if existing state is at least Hibernate? I really doubt someone would accidentally have that setting

IceStormNG:
We could do that if you want to. So, if the value is 2 or 3 we save it if the option does not exist yet in the config.json.

seerge:
My point here is to keep app non invasive as possible:)

IceStormNG:
Ok. I added code to read and save the setting on startup of ghelper if it is either hibernate or shutdown.
Sleep will not be saved as it is the default if uncofigured.
If it is set to "Do nothing" it is assumed that it was set by GHelper. If the user did that themselves, they shouldn't use the clamshell mode toggle.

seerge:
@IceStormNG great, thanks!

PR #907, opened 2023-07-27
ASUS Mouse: Resolved deadlocks

IceStormNG:
The deadlocks that happen with the mouse provider due to recent refactoring are now resolved.

I also re-enabled device polling during "RefreshSensors". The reason is, that this is also used to determine whether the mouse is ready for setting changes. The mouse sends no event when it gets turned off. The dongle still stays there and no HID device event is raised. You only know by reading the mouse and finding out that the mouse responds with empty packets.
The performance impact should be negligible as it runs on a separate thread and even that only takes a few ms.

I also changed the calls to "RefreshBatteryForAllDevices" to run in async tasks as the constant USB traffic for refreshing should still not run on main thread. It could lock the UI if the device is slow to respond.

Also fixed a crash if you close the window while the battery gets updated.

### DIFF
--- a/app/AsusMouseSettings.cs
+++ b/app/AsusMouseSettings.cs
@@ -82,7 +82,7 @@ namespace GHelper
 
             InitMouseCapabilities();
             Logger.WriteLine(mouse.GetDisplayName() + " (GUI): Initialized capabilities. Synchronizing mouse data");
-            Task task = Task.Run((Action)RefreshMouseData);
+            RefreshMouseData();
         }
 
         private void AsusMouseSettings_FormClosing(object? sender, FormClosingEventArgs e)
--- a/app/AsusMouseSettings.cs
+++ b/app/AsusMouseSettings.cs
@@ -94,14 +94,14 @@ namespace GHelper
 
         private void Mouse_MouseReadyChanged(object? sender, EventArgs e)
         {
+            if (Disposing || IsDisposed)
+            {
+                return;
+            }
             if (!mouse.IsDeviceReady)
             {
                 this.Invoke(delegate
                 {
-                    if (Disposing || IsDisposed)
-                    {
-                        return;
-                    }
                     Close();
                 });
             }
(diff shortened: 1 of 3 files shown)


seerge:
@IceStormNG small thing i noticed,

If connected to a dock when laptop goes to sleep / wakes up , it will obviously get a bunch of USB connection events on wake up from each peripheral

So in a log it looks like this 

```
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
7/27/2023 7:33:00 PM: HID Device Event: Checking for new ASUS Mice
```

IceStormNG:
Doesn’t look too bad. It checks a few times. Once per device detection. 

That should not cause any issues

seerge:
@IceStormNG oke, I just noticed that memory usage by app jumps significantly after waking up (you can check yourself in task manager)

I tried to debug it (even tho i'm not really good at that), by doing 2 memory dumps before / after short sleep

Main increase was thanks to HidSharp objects

Screenshot 2023-07-27 203308

seerge:
PS mb i look at wrong numbers tho :) i’m just trying to figure out reasons why it eats more memory

IceStormNG:
My code only creates new HID devices when it finds a supported one to connect to. Otherwise, it just checks for the device IDs but doesn't open streams.
The objects should get disposed soon.

Issue #930, opened 2023-07-30
[bug] Flow Z13 Rear Window Lighting break since 0.97 update.


rainbowflesh:
Follow the #843, I test all release from 0.89 to 0.100, and located .97 will make `Rear Window Lighting` break on ghelper.exe start, but can be re-enabled in Armoury Crate.

I'm trying to find commits that contain this bug, I can physically access to the device, and I'm also SoC software developer so if anything I can help I'll try my best ;)

**Describe the bug**
Z13 Rear Window Lighting (hereinafter called *RWL*) break since 0.97 update.

**Clear scenario to Reproduce**
1. Turn on RWL in Armoury Crate.
2. run 0.96, it okey nothing change.
3. run 0.97 RWL will turn off.

**Expected behavior**
Like 0.96, runing ghelper won't turn off RWL.

**App Logs**
[97_log.txt](https://github.com/seerge/g-helper/files/12208545/97_log.txt)
[96_log.txt](https://github.com/seerge/g-helper/files/12208546/96_log.txt)
```git diff
# git diff 096.log 097.log
+ 2023/7/30 13:06:50: TL134ADXP01-0:DISPLAYCONFIG_OUTPUT_TECHNOLOGY_INTERNAL: \\.\DISPLAY1
+ 2023/7/30 13:06:52: USB-KB 0x18C6:5D-BD-01-FF-1E-0F
+ 2023/7/30 13:06:52: Auto:5D-BA-C5-C4-03
+ 2023/7/30 13:06:52: USB-KB 0x1A30:5D-BD-01-FF-1E-0F
- 2023/7/30 13:06:30: Auto:5D-BA-C5-C4-03
```

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Laptop model: z13

**Asus software**
- Armoury crate (or it's services installed)
- Only `AsusOptimization.exe` except Armoury Crate stuffs.

----
edit:
most suspicious commit https://github.com/seerge/g-helper/commit/e5f0d77d05f0388b5449f7361897a60aaf54a4c5

rainbowflesh:
I build debug on commit `3ffcc53b` and everything works fine, then build `e5f0d77d05f0388b5449f7361897a60aaf54a4c5` the RWL broke 😸 seems buggy founded.

rainbowflesh:
By comment function `AsusUSB.ApplyAuraPower()` make the light work again, which implement in `app/AsusUSB.cs`, l: 311

```csharp
[...]

PR #933, opened 2023-07-30
Added support for ASUS TUF Gaming M4 Wireless (P306_Wireless) Mouse

IceStormNG:
TUF Gaming M4 Wireless (P306_Wireless) is now supported (thanks to @alexatthehub for testing and providing data) 
This also adds some general fixes for the mouse system. 
Battery is now only polled every 20s to reduce useless USB traffic, except for when the GHelper window opens. Then it is force-refreshed. Battery polling now happens in background thread to not block the UI.
Added better error handling when the mouse sends unexpected packets.

Some internal housekeeping to allow for easier crafting of new mouse classes with less redundant code.
Fixes the DPI colors on the buttons for mice that have no custom DPI colors in hardware.

### DIFF
--- a/app/AsusMouseSettings.cs
+++ b/app/AsusMouseSettings.cs
@@ -473,6 +473,8 @@ namespace GHelper
 
             if (mouse.HasRGB())
             {
+                sliderBrightness.Max = mouse.MaxBrightness();
+
                 foreach (LightingMode lm in Enum.GetValues(typeof(LightingMode)))
                 {
                     if (mouse.IsLightingModeSupported(lm))
--- a/app/AsusMouseSettings.cs
+++ b/app/AsusMouseSettings.cs
@@ -391,9 +391,9 @@ namespace GHelper
                 buttonDPI4.Image = ControlHelper.TintImage(Properties.Resources.lighting_dot_24, Color.Green);
 
                 buttonDPI1.BorderColor = Color.Red;
-                buttonDPI1.BorderColor = Color.Purple;
-                buttonDPI1.BorderColor = Color.Blue;
-                buttonDPI1.BorderColor = Color.Green;
+                buttonDPI2.BorderColor = Color.Purple;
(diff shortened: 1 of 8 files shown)


seerge:
@IceStormNG great, thanks a lot again for contribution.

How do you think, is this implementation stable enough to go "live" in a "latest release" state?

My current release strategy, is to once per last 3-4 releases make one "final" (as soon as it gathers ~ 1000 downloads from enthusiasts). This will push it to all remaining users via auto-update feature (as it checks only for "final" ones).

IceStormNG:
I'm currently finalizing another mouse and RGB zoning. Not sure whether you want to wait for that but I will likely submit a pull request in a few hours for that.

As for stability. I run the latest build on my machine all the time and have fixed all bugs that I encountered. I guess there are not that many users with ASUS mice that are active here to submit bug reports.

I personally would like to add the Keris mouse and the RGB zoning before release if that is ok.

seerge:
@IceStormNG i'm absolutely fine to wait :) 

+ I still want to fix this thing https://github.com/seerge/g-helper/issues/930 (need to wait for a user to respond)

When 103 will go live, i usually wait for a couple of days to gain ~ 1000 downloads or so ( unless someone spots a new issue there ) and then push it to autoupdate (otherwise windows defender may start flagging it if it )

IceStormNG:
image


Btw. I made a new screenshot with the new zone buttons if you want to update that.

I'm waiting for one person to respond for testing the last changes to the Keris. Then you can release another version if you like to. The other mice will then come with a later version. There are already 2 new requests :D

seerge:
@IceStormNG there is no rush with release :)

IceStormNG:
That's good. Then I might also get the TUF M3 and the Strix Impact II Wireless also ready if the requestors tested and responded. :)

It's interesting. Some mice have more features in their firmware than AC lets you use. I enable those in GHelper as well if they work 😆

PR #938, opened 2023-07-30
Added Rog Keris Wireless

IceStormNG:
This adds the ROG Keris wireless with _almost_ full support.

One issue is here: You cannot switch the DPI profile. You can do it via the mouse buttons but the mouse does not respond on the AC interface for DPI switching via GHelper. We tested a lot and haven't found any clue how AC does it. There is 0 traffic over the USB Bus that indicates a DPI profile change. The default endpoints, that even work on mice without AC support for DPI switching (like the TUF M4) error out on this mouse (mouse responds with FF AA, which seems to be a general "I don't understand that packet" message). Either it is a firmware bug or ASUS did something goofy here.

We might have to document that for the mouse support. Everything else works fine for this mouse as tested by @1m4g1ne.

### DIFF
--- a/app/Peripherals/Mouse/Models/ROGKerisWireless.cs
+++ b/app/Peripherals/Mouse/Models/ROGKerisWireless.cs
@@ -1,0 +1,169 @@
+﻿namespace GHelper.Peripherals.Mouse.Models
+{
+    //P513
+    public class ROGKerisWireless : AsusMouse
+    {
+        public ROGKerisWireless() : base(0x0B05, 0x1960, "mi_00", true)
+        {
+        }
+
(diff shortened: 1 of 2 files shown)


seerge:
@IceStormNG great, is it final PR, or you still want to add something in a nearest time ?

IceStormNG:
I have 2 mice almost ready. I just wait for response from the testers. But I cannot tell a) when they respond and b) how often I have to fix something.
So you can already release it as it is for now and I can add those later.

Btw: I just also updated the German translations on crowdin

seerge:
@IceStormNG great,

It's live https://github.com/seerge/g-helper/releases/tag/v0.103 (as pre-release) :)

PR #977, opened 2023-08-03
add as much button support as we can to the RC71 (ROG Ally)

ktemkin:
Just a quick patch I made for myself (and a quick fix for custom commands using the M2 button) -- thought I'd send this upstream in case it's helpful. :)

### DIFF
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -45,7 +45,7 @@ namespace GHelper
             comboFNE = new RComboBox();
             textFNE = new TextBox();
             labelFNC = new Label();
-            textBox1 = new TextBox();
+            textM2 = new TextBox();
             textM1 = new TextBox();
             comboM1 = new RComboBox();
             labelM1 = new Label();
@@ -63,7 +63,6 @@ namespace GHelper
             comboFNC = new RComboBox();
             textFNC = new TextBox();
             tableKeys = new TableLayoutPanel();
-            textM2 = new TextBox();
             panelBacklightHeader = new Panel();
             sliderBrightness = new Slider();
             pictureBacklight = new PictureBox();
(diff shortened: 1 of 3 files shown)


seerge:
@ktemkin hello, thanks for contribution, that can work I guess :)

k-trahan:
I don't think this is quite right.

"m1" is still considered the volume down button on the Ally and not the back paddles. "m2" is still volume up and by hiding the label and combo box, you lose the ability to adjust the volume up hotkey.

At a minimum, I think further work is needed in Input/InputDispatcher.cs around line 224 for special handling of RC71 hardware. Also need additional labels and inputs for the two volume buttons.

seerge:
@k-trahan did you update to 0.105?

k-trahan:
> @k-trahan did you update to 0.105?

Yes, this is after updating to v0.105.

You can observe some of this by using 0.103, changing the volume down hotkey, changing back to 0.105 and noticing that your hotkey you just set is defined in "Back Paddles".

ktemkin:
I see what's happening -- the hooks are still there for the Vol+/Vol- translation into M1/M2 in a code path targeted for other devices.

I'll have a fix in a sec. >.<

seerge:
@k-trahan @ktemkin 
[GHelper.zip](https://github.com/seerge/g-helper/files/12264072/GHelper.zip)

This one should skip binding to volume keys :) + i have reorganized some stuff inside

ktemkin:
https://github.com/seerge/g-helper/pull/1003 <-- I threw this PR up as a quick fix, but if you have a better version than my quick two-line tweak, feel free to just close it. :)

k-trahan:
Tested the build above and it looks good to me, Appreciate you both!

seerge:
@ktemkin I'm sorry to brutally intrude into your PR, but i have made some adjustments (in main branch). So now Ctrl Center + Paddle has own param name (for config) + will have default action as nothing (-----).

As by just reusing g-helpers M1, M3 it would just bind volume-down and mic-mute to it otherwise, which may confuse people who just installed it

I didn't touch Ctrl+M1/M2 and Shift + M1/M2 bindings, as (theoretically), why not to keep them for ally? and control brightness with Ctrl + volume keys? :))) i personally missed that on my ally

Build with updates : 
[GHelper.zip](https://github.com/seerge/g-helper/files/12264232/GHelper.zip)

k-trahan:
> @ktemkin I'm sorry to brutally intrude into your PR, but i have made some adjustments (in main branch). So now Ctrl Center + Paddle has own param name (for config) + will have default action as nothing (-----).
> 
> As by just reusing g-helpers M1, M3 it would just bind volume-down and mic-mute to it otherwise, which may confuse people who just installed it
> 
> I didn't touch Ctrl+M1/M2 and Shift + M1/M2 bindings, as (theoretically), why not to keep them for ally? and control brightness with Ctrl + volume keys? :))) i personally missed that on my ally
> 
> Build with updates **: [GHelper.zip](https://github.com/seerge/g-helper/files/12264265/GHelper.zip)
> 
> I hope it works (i have returned my Ally back to the store, so can't really test anymore)

This build works, too, and I like the change with the default action being "-----".

I also **really** like the idea of adjusting brightness with CTRL + M1/Vol. Down and CTRL + M2/Vol. Up. I'm probably overlooking something obvious but how can I set these hotkeys? On my Ally, the only Key Bindings I have now are Ctrl Center, ROG, and Back Paddles.

(Also bummed that you no longer have your Ally, @seerge!)

seerge:
@k-trahan i think you can bind some of gamepad keys (like RB or RT to be control or shift?) You can also try to bind back-paddle to be Ctrl (via custom hotkeys in g-helper itself), but main problem here - is app gets only CLICK event for padddle/ctrl center/rog (so it doesn't know if button is still pressed) and therefore it can't emit holding Control, only clicking it in turn.

k-trahan:
Makes sense, thanks so much for the explanation, @seerge.

PR #1010, opened 2023-08-05
Support for rog keris eva edition mouse

Rakaveli:
Hi :) added support for my rog keris evangelion edition mouse. As it turned out, it has different id in contrast to the standard keris mouse.


seerge:
@Rakaveli oke, thanks!

PR #994, opened 2023-08-04
Updated README.md for better readability.

kindaLakshay:
Punctuation and grammar changes without changing the formatting.

PR #1019, opened 2023-08-07
Update README.zh-CN.md

MrChenYukun:
1. update FAQ section's chinese translation from README.md to README.zh-CN.md
2. unify symbols ( （ -> ( , ） -> ) )

### DIFF
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,3 +1,4 @@
+
 # G-Helper (GHelper)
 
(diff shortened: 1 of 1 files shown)


seerge:
@MrChenYukun hello, thanks for addition 

As I remember @sklynic was helping with initial translation, does it make sense to completely re-translate whole readme (as it has changed quite a bit since that moment) or its fine as is?

How is automatic google translation quality for such things in Chinese? mb it just makes sense to have auto-translated link instead of editing it all the time?

MrChenYukun:
@seerge Hi, re-translate whole readme is quiet a lot work, so i just translated this section.
I see you mentioned that you are reorganizing the README.md file(https://github.com/seerge/g-helper/pull/994)? So i don't know if it's a good time to re-tranlate whole README.md now.
Google translation quality is ok, but since Chinese is quite a different from English, certain complex sentences or those with semantic context, there may be some inaccuracies or lack of fluency in the translation. For more precise and natural translation results, it is still recommended to use human translation.

PR #1096, opened 2023-08-15
AniMeMatrix fix for GU604

IceStormNG:
Back when I made the AniMeMatrix Support for the GU604 (Zephyrus M16 2023), I somehow screwed up the first 10 rows that were shifted by 1 LED in x direction.
This is fixed now.
 
The frameShift that I introduced back then is not needed anymore, I just had to account for every other row having different amounts of LEDs for the full rows.

I have now tested it with some stripe patterns to make sure they're all straight and not jagged or have shifts.

### DIFF
--- a/app/AnimeMatrix/AnimeMatrixDevice.cs
+++ b/app/AnimeMatrix/AnimeMatrixDevice.cs
@@ -90,8 +90,6 @@ namespace Starlight.AnimeMatrix
         //public int FullEvenRows = -1;
 
         public int dx = 0;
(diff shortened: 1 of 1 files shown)


seerge:
Thanks! I suspected that :)

IceStormNG:
Yeah. It looked good at first, but I only tested it with some GIFs. Now with the time, the upper numbers looked off so I made some images to test it properly. the Matrix was a little more complicated than the other ones implemented.

PR #1117, opened 2023-08-18
Add Power Limit for ROG Ally & fix Designer

patrickpielarski:
- add Power Limit for ROG Ally 
- fix Designer for Forms not working in Visual Studio

### DIFF
--- a/app/AsusACPI.cs
+++ b/app/AsusACPI.cs
@@ -215,6 +215,7 @@ public class AsusACPI
         }
 
         if (AppConfig.IsAdvantageEdition()) MaxTotal = 250;
+        if (AppConfig.IsAlly()) MaxTotal = 30;
         if (AppConfig.IsX13())
         {
             MaxTotal = 75;
--- a/app/Updates.cs
+++ b/app/Updates.cs
@@ -6,16 +6,6 @@ using System.Text.Json;
 
 namespace GHelper
 {
-
-    public struct DriverDownload
-    {
-        public string categoryName;
-        public string title;
-        public string version;
-        public string downloadUrl;
-        public JsonElement hardwares;
-    }
-
     public partial class Updates : RForm
     {
         //static int rowCount = 0;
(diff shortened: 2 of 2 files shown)


seerge:
@patrickpielarski hello, but Ally supports 43W+ even :) higher values most probably won't have much effect (as there is a hard limit in BIOS)

 I can add a limit to 43W or 50W if you want

patrickpielarski:
@seerge  Hi, didn't know about it being able to go higher than 30W. But with 43W the CPU voltage should be dropped also? If not the Ally will throttle itself back to 30W after 2 minutes according to:
https://www.reddit.com/r/ROGAlly/comments/14iqv30/4243w_turbo_mode_gpu_throttle/
https://www.reddit.com/r/ROGAlly/comments/14v5h88/43w_rog_ally_gives_much_smoother_experience/

But a value lower than 150 like 43 or 50 would be appreciated. Just don't know if it really works.

seerge:
@patrickpielarski Ally has sustained, long and short boosts.

It can go up to 43W officially even. So I think 50W would be ok (mb they will unlock something in future via BIOS)

I have added that https://github.com/seerge/g-helper/commit/3652e4d0ffce2b96c00f64b14d29eb6011d60000

patrickpielarski:
> @patrickpielarski Ally has sustained, long and short boosts.
> 
> It can go up to 43W officially even. So I think 50W would be ok (mb they will unlock something in future via BIOS)
> 
> I have added that [3652e4d](https://github.com/seerge/g-helper/commit/3652e4d0ffce2b96c00f64b14d29eb6011d60000)

@seerge Thanks! The other change in my branch was putting the struct away from top of the file so Visual Studio opens the Designers again for the Forms.

patrickpielarski:
> @patrickpielarski hmm, my studio (latest version) doesn't have problems with that, but I moved struct inside form, so it should be ok now
> 
> [584755e](https://github.com/seerge/g-helper/commit/584755ea93aca8e51a03c5f51608e25eb2eaceea) :)

Thanks. Just installed 2022 Community Edition on the Ally to test it and it wouldn't open the designers because of that one struct. Appreciated!

seerge:
Do you mean it still doesn't work for you? :)

patrickpielarski:
No it works now. I was just claryfying that it wasn't working before. Thanks

PR #1120, opened 2023-08-18
Internal improvements for mouse support

IceStormNG:
This is some internal improvement for further mice and also to make extensibility easier. The current mice make no use of it yet, but the Harpe Ace, which is almost done, will be the first (and maybe only one) that needs all those extra stuff.

It also improves error handling and out-of-band or empty packets that get send by some mice here and there.


seerge:
@IceStormNG great, thanks for your dedication again :)

Issue #1092, opened 2023-08-15
G-Helper using Windows mode instead of App mode for theme


ct16k:
I have Windows configured on Dark mode but Apps configured to Light mode, as this is the only combination that will allow me to customize the start menu color with light a mode theme. In this setup, G-Helper app renders with a dark theme instead of light. If I switch the two settings around (i.e. Windows to Light and Apps to Dark), G-Helper will render with a light theme, confirming it's looking at the Windows mode, not app mode, to decide on the theme color.

**Clear scenario to Reproduce**
1. Go to Windows Settings / Personalization / Colors
2. Set default Windows mode to Dark
3. Set default app mode to Light
4. G-Helper is rendered with a dark theme, not light

**Expected behavior**
G-Helper to be rendered with a light theme (or whatever app mode is set to)

**App Logs**
[log.txt](https://github.com/seerge/g-helper/files/12342903/log.txt)

**Screenshots or screencasts**
![image](https://github.com/seerge/g-helper/assets/7895891/6318781b-038f-43d0-86e4-d212400069d1)
![image](https://github.com/seerge/g-helper/assets/7895891/fd677b3c-58fd-4f7e-b2d1-66092ee3a880)

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - GV302XI

PR #1125, opened 2023-08-18
Dark/Light Mode override

IceStormNG:
With the recent changes, my setting (dark taskbar, light window) results in light GHelper which was dark before.

So, I added an option to override the light/dark mode detection in the config file.

The setting is called `ui_mode` and can be either `dark `or `light` (case-insensitive) to override the default design. Everything else, or omitting the setting, will result in the current behavior, which is auto detection.

### DIFF
--- a/app/UI/CustomControls.cs
+++ b/app/UI/CustomControls.cs
@@ -62,6 +62,16 @@ namespace GHelper.UI
 
         private static bool IsDarkTheme()
         {
+            if (AppConfig.GetString("ui_mode").ToLower() == "dark")
+            {
+                return true;
+            }
+
+            if (AppConfig.GetString("ui_mode").ToLower() == "light")
+            {
+                return false;
+            }
+
             using var key = Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\Windows\CurrentVersion\Themes\Personalize");
             var registryValueObject = key?.GetValue("AppsUseLightTheme");
 
--- a/app/UI/CustomControls.cs
+++ b/app/UI/CustomControls.cs
@@ -62,12 +62,14 @@ namespace GHelper.UI
 
         private static bool IsDarkTheme()
         {
-            if (AppConfig.GetString("ui_mode").ToLower() == "dark")
+            string? uiMode = AppConfig.GetString("ui_mode");
+
+            if (uiMode is not null && uiMode.ToLower() == "dark")
             {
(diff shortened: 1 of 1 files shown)


seerge:
@IceStormNG oke :) so you actually want literally opposite to what user from this topic asked https://github.com/seerge/g-helper/issues/1092

IceStormNG:
> @IceStormNG oke :) so you actually want literally opposite to what user from this topic asked #1092

Yes. I preferred the older behavior. So, this way I can set an override to my likings :D

seerge:
@IceStormNG btw, since you have nvidia / optimus and external monitor -  can you quickly check if new "internal screen detection" https://github.com/seerge/g-helper/commit/190786a15a7ae94f81da0df8e6a12b3373ba68c6 works in your scenarios as well ? (i.e. compile app and test that a bit) :)

IceStormNG:
It works when the internal screen is disabled.

image

And works fine when the internal is enabled + external.

image


Only issue I just notice... with two screens, GHelper positions itself wrongly. The left is the laptop screen (primary), right one is the external (secondary). GHelper launches exactly where you see it.

image

seerge:
@IceStormNG thanks :) 

positioning is something else, coordinate calculation is launched every time window gets shown, but apparently windows itself is not always consistent with what it reports as screen size … :)

AudreyAP:
Just an idea - maybe instead of specifying the specific color mode, also allow specifying "app" or "windows" to select which theme the app follows? For me personally ghelper is more like the windows menu than an app, makes more sense to follow windows settings. Others will disagree!

seerge:
@AudreyAP well, it was windows before - now it's app, cause someone has asked. I'm not a big fan of adding million of checkboxes in the app :) but now there is an option to set it via config :)

AudreyAP:
Doesn't need to be a checkbox in the app, config option alone to select following windows would be great - problem is now that I need to open the config twice a day to make it match my start menu haha

IceStormNG:
> Doesn't need to be a checkbox in the app, config option alone to select following windows would be great - problem is now that I need to open the config twice a day to make it match my start menu haha

I opened a pull request that allows for that. So you will have to set  `windows` for ´ui_mode` in the config file with the new change (once merged).

AudreyAP:
@IceStormNG thank you so much!

PR #1137, opened 2023-08-20
Support for ROG Harpe Ace Aim Lab Edition (P713).

IceStormNG:
Mouse support for the ROG Harpe Ace Aim Lab Edition (P713). Mouse is supported wired, via wireless dongle and also through an OMNI Receiver.

### DIFF
--- a/app/Peripherals/Mouse/AsusMouse.cs
+++ b/app/Peripherals/Mouse/AsusMouse.cs
@@ -141,7 +141,6 @@ namespace GHelper.Peripherals.Mouse
     {
         private static string[] POLLING_RATES = { "125 Hz", "250 Hz", "500 Hz", "1000 Hz", "2000 Hz", "4000 Hz", "8000 Hz", "16000 Hz" };
         internal const bool PACKET_LOGGER_ALWAYS_ON = false;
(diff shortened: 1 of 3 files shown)


seerge:
@IceStormNG great! Congrats on finishing it.

What was the main reason it didn’t work? A custom packet size?

IceStormNG:
Thanks

Yes and also something else. Yes, normally the mouse wants 64bytes packets (+1 for the report id). But through the omni receiver it only wants 63 bytes (+ 1 for the report id, which is not 0 here, like for all other mice, but 3).
And: it has multiple sub devices. Only one of them has the report id 3 and support interrupt write and read, which is the one that talks to the mouse.

seerge:
@IceStormNG i see, XGM has a similar trick, it wants 300 bytes of data as it seems

And it’s indeed unrealistic to catch all that w/o device :)

IceStormNG:
Yeah. I hope I don't have to "abuse" amazon return policy more often or they might not like that. Most mice got implemented without trickery, though I could use one of the mice where the DPI switch doesn't work. I guess they send the DPI switch command through another device endpoint or with a different report ID.

PR #1153, opened 2023-08-22
Fix default fan curve buffer size

charlespeng:
When getting default fan curve by GetFanCurve(), CallMethod() returns buffer with size 20, and causes IsInvalidCurve() fail.
It seems other functions only use the first byte of returned buffer; we can safely reduce the buffer size to 16 to fix default fan curve.


seerge:
@charlespeng probably missed that while refactoring something. Thanks for spotting

PR #1167, opened 2023-08-23
New binding to previous performance mode (CTRL + SHIFT + F21)

hotbarlo:
The only thing this does is create a new binding to previous performance mode. Reason was that the ROG Ally doesn't have a FN key. I'd like to create more bindings to F13-F24 keys so that Ally users have more options to configure G-Helper. Let me know how you feel about it!

### DIFF
--- a/app/Input/InputDispatcher.cs
+++ b/app/Input/InputDispatcher.cs
@@ -14,6 +14,7 @@ namespace GHelper.Input
         public static bool backlightActivity = true;
 
         public static Keys keyProfile = Keys.F5;
+        public static Keys keyProfilePrevious = Keys.F21;
         public static Keys keyApp = Keys.F12;
 
         static ModeControl modeControl = Program.modeControl;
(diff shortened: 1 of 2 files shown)


seerge:
@hotbarlo hello, thanks for contribution.

Currently ``Ctrl + Shift + F5`` hotkey is even configurable via ``config.json``. So I prefer to keep this logic.

I have added ``Ctrl + Shift + Alt +F5`` as combination to cycle via modes backwards (same as it happens with FN+F5).
https://github.com/seerge/g-helper/commit/47f86d222f6b4f97bcfb550b6f558f1501e51404#diff-3c45dfef8460c8a53dcc57e2fe4a4c0c9145c72878df616ca1beb1c76242996f

You can check this build
[GHelper.zip](https://github.com/seerge/g-helper/files/12428088/GHelper.zip)

seerge:
@hotbarlo 

Change is live at https://github.com/seerge/g-helper/releases/tag/v0.116 :)

Closing this PR cause it's implemented in a bit different way already.

PR #1174, opened 2023-08-24
Multi Screen position & DPI Optimize

kingsznhone:
# Blur Fix 
I use `<dpiAwareness>PerMonitorV2</dpiAwareness>` to replace `<dpiAwareness>System</dpiAwareness>` awareness
[https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process](url)

This fix blur text and icon while screens have diffirent DPI (150% for 1st screen, 125% for 2nd screen on my computer) 

Also effect on every sub form. 

### Old 
Old

### New
New

Image is compressed by github. The visual experience is better on native machine.

# Multi Screen Drag Fix

While using `<dpiAwareness>System</dpiAwareness>`, form will not resize when crossing screen.

But under `<dpiAwareness>PerMonitorV2</dpiAwareness>` , DWM will resize form 5 times to find a better render size.

this will trigger `settingForm.Resize()`. Force Form position back to right bottom cornor at initial screen. 

I make Resize() only working on crossing screen. Then relocate form to right bottom cornor of another screen by canculating screen offset.

### DIFF
--- a/app/Program.cs
+++ b/app/Program.cs
@@ -217,15 +217,18 @@ namespace GHelper
             if (settingsForm.Visible) settingsForm.HideAll();
             else
             {
+                Rectangle CurrentScreen = Screen.FromControl(settingsForm).WorkingArea;
+                Point CurrentScreenOffset = Screen.FromControl(settingsForm).WorkingArea.Location;
 
-                settingsForm.Left = Screen.FromControl(settingsForm).WorkingArea.Width - 10 - settingsForm.Width;
-                settingsForm.Top = Screen.FromControl(settingsForm).WorkingArea.Height - 10 - settingsForm.Height;
+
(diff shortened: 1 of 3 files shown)


kingsznhone:
Is there any plan for moving to  WPF or winUi3？ 
To be honest, winform is too old to use. 
For example, v2rayN , a widely used winform .net framework network tool,  after years of struggle, finally  move to wpf .net6 this year.

wertone-gh:
You could moving córę functionality to Windows servcie and use UI ghelper just a as a client. This will be even more efficient

seerge:
@wertone-gh i can, but that won’t make app lighter and taking less resources:) And would take a lot of time

PR #1264, opened 2023-09-06
Anime Matrix tweaks

seerge:
Fixed matrix rendering for G14 2020/2021 and added new Clock design :)

### DIFF
--- a/app/AnimeMatrix/AnimeMatrixDevice.cs
+++ b/app/AnimeMatrix/AnimeMatrixDevice.cs
@@ -4,7 +4,6 @@ using GHelper.AnimeMatrix.Communication;
 using System.Drawing.Drawing2D;
 using System.Drawing.Imaging;
 using System.Drawing.Text;
-using System.Globalization;
 using System.Management;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


seerge:
@IceStormNG hello :)

Thanks to your last PR or Matrix on M16 I have finally figured out and properly fixed matrix for G14 2020/2021 (it's the worst version of it)

I have also improved clock mode rendering (including native matrix font) and added date to it.

Can you check if it looks ok on your M16 screen ? and if not - set correct value here for your sizes ?
https://github.com/seerge/g-helper/blob/c791421c3e328d27361a055b834605681663c652/app/AnimeMatrix/AnimeMatrixDevice.cs#L131

seerge:
@IceStormNG i didn't really center it , but tried to fit into smallest version of matrix on G14 2021 (it has least amount of leds)

IceStormNG:
IMG_1765

This is how it looks with the current state without any change by me

seerge:
@IceStormNG i think it's fine as is then

I may try to do "proper" centering later, but then I need to calculate actual length of the diagonal (taking into account the height of the text itself as well)

But for now it will work :)

PR #1277, opened 2023-09-08
Fan calibration

seerge:
Added a button to "Calibrate" Fan Axis (i.e. measure maximum RPM) for each fan, and adjust axis accordingly.

Actual fan behavior isn't affected anyhow, as it's set in %

### DIFF
--- a/app/AsusACPI.cs
+++ b/app/AsusACPI.cs
@@ -1,4 +1,5 @@
-﻿using System.Management;
+﻿using GHelper;
+using System.Management;
 using System.Runtime.InteropServices;
 
 public enum AsusFan
@@ -346,6 +347,32 @@ public class AsusACPI
         return -1;
     }
 
+    public int GetFan(AsusFan device)
+    {
+        int fan = -1;
+
+        switch (device)
+        {
+            case AsusFan.GPU:
+                fan = Program.acpi.DeviceGet(GPU_Fan);
+                break;
+            case AsusFan.Mid:
+                fan = Program.acpi.DeviceGet(Mid_Fan);
+                break;
+            default:
+                fan = Program.acpi.DeviceGet(CPU_Fan);
+                break;
+        }
+
+        if (fan < 0)
+        {
+            fan += 65536;
+            if (fan <= 0 || fan > 100) fan = -1;
+        }
+
+        return fan;
+    }
+
     public int SetFanRange(AsusFan device, byte[] curve)
     {
         byte min = (byte)(curve[8] * 255 / 100);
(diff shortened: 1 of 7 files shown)


ke1ne:
Cool stuff! I would add some warning, like "This will drive your fans to the maximum speed! Also, can be loud."
Also, the CANCEL button with smooth slowing down could be grate too.
Thank you!

seerge:
@ke1ne well, not louder than Turbo in most cases :)

Setting any mode will "cancel" that (as it will set original fan curve back)

But app will do that when calibration is finished anyway (i.e. when it will see that fan speed is not raising anymore)

ke1ne:
@seerge I never got those speeds before, even with Turbo and active gaming. It feeled almost almost like "fans are gonna break" (6000+ rpm).

seerge:
@ke1ne lucky you. It just sets 100% fan curve, that's it :) it won't break anything

ke1ne:
@seerge OK, got it. Thanks!

seerge:
@tdj88 currently ``Calibrate`` button appears for device where it's possible to read default fan curve. Both G14 and G16 2024 just respond with zeroes instead of one, so app assumes that's not supported and hides a button :)

The only thing Calibrate feature does - it sets custom fan curve with all dots at 100% and then waits until fan speed stops at some value, to remember it adjust Y-axis on a chart with maximum recorded RPM. But same happens if app just spots higher RPM during normal operation. It remembers that.

Problem is that both G14/G16 2024 have insanely slow reaction speed on fans, it took me like 2 minutes until fans reached their max RPM.

What you can do - is set curve to 100% manually and just write me your maximum values. I can hardcode them as defaults for your model  here
https://github.com/seerge/g-helper/blob/300eb7af7027df4030c568fea63f0cb4d2372044/app/Fan/FanSensorControl.cs#L46

tdj88:
@seerge 
CPU MAX SPEED= 6100 RPM (goes between 6100 and 6200)
GPU MAX SPEED= 6100 RPM (goes between 6100 and 6200)
MID FAN MAX SPEED = 9200 RPM

seerge:
@tdj88 ok, great. Is it based on Turbo mode under the hood ?

tdj88:
@seerge I just did balance  underhood for max fan speed

seerge:
@tdj88 can you check if it's same with Turbo?

tdj88:
@seerge Same speeds in Turbo  (Note: Mid fan went up to 9300  for  restricted  airflow, lifting the laptop is drops to 9100 to 9200 RPM)

seerge:
@tdj88 ok, added

This build should show calibrated values from start

[GHelper.zip](https://github.com/seerge/g-helper/files/14998301/GHelper.zip)

PR #1282, opened 2023-09-09
Auto close asus apps

gloatoriginal:
Added a setting in the extras menu to auto close asus apps upon ghelper launch.


seerge:
@gloatoriginal hello, thanks for contribution

But what is the practical purpose of that ? When you "Stop" services in a first place, it would set them to Disabled which means they won't start again next time anyway.

The only (rare) exception, is when you would update "Asus System Control Interface" driver, and it would enable services again.

App UI is already overloaded with different checkboxes, that make it more confusing to use - so I'm trying to keep things simple :|

gloatoriginal:
Hey thanks for looking into my contribution, I greatly appreciate your time in doing so!

I apologize for having not noticed the disabled functionality on the asus services. Currently on my Zephyrus G14 2022 model it starts ~8 services on reboot/start up, which I would than manually shut off via ghelper -> Extras each time, so I wanted to circumvent that by having them just auto-close on startup.

However when I implemented this auto shut off asus apps I believe it was on version ~.118 so I'm curious if the disabling came later, I will rebase if that is the case because you are correct that'd make this functionality I provided obsolete/not useful. 

Thank you for the time and consideration!

PR #1372, opened 2023-09-27
Updated Turkish Translation (Strings.tr.resx)

mertatilgan:
Here is the Turkish translation that is compatible with the current version of G-Helper. There may be some minor translation errors due to localisation of some terms, but for the most part everything should be fine.


seerge:
@mertatilgan hello, thanks for your contribution

But for a long time I do [all translations via Crowdin](https://github.com/seerge/g-helper/discussions/759)

Can you sign up to Crowdin and do everything there ? As if you push it in manual way, it screws up sync often :)

seerge:
@mertatilgan  I have merged your additions from Crowdin, and they will be included in next release :)

Therefore closing this PR :)

Thanks again!

PR #1407, opened 2023-10-02
feat: improve focus behavior

nopeless:
Before: when g helper window is out of focus/behind a window, pressing the ROG key will make it disappear, thus making the user have to press twice. Additionally, when clicking the taskbar icon, it only focuses on the settings window. This causes issues when you click other sub panel icons as it hides a non-visible window, confusing the user.

After: pressing ROG key will bring the window to top if it is not in focus. Clicking taskbar icon or activating ROG key will focus all previously visible subpanels to top.

### DIFF
--- a/app/Input/InputDispatcher.cs
+++ b/app/Input/InputDispatcher.cs
@@ -337,7 +337,7 @@ namespace GHelper.Input
             if (e.Modifier == (ModifierKeys.Control | ModifierKeys.Shift))
             {
                 if (e.Key == keyProfile) modeControl.CyclePerformanceMode();
-                if (e.Key == keyApp) Program.SettingsToggle();
+                if (e.Key == keyApp) Program.SettingsToggle("", true);
                 if (e.Key == Keys.F20) KeyProcess("m3");
             }
 
@@ -427,7 +427,7 @@ namespace GHelper.Input
                     {
                         Program.settingsForm.BeginInvoke(delegate
                         {
-                            Program.SettingsToggle();
+                            Program.SettingsToggle("", true);
                         });
                     }
                     catch (Exception ex)
--- a/app/Program.cs
+++ b/app/Program.cs
@@ -213,11 +213,21 @@ namespace GHelper
             SetAutoModes(true);
         }
 
-
-
         public static void SettingsToggle(string action = "")
         {
-            if (settingsForm.Visible) settingsForm.HideAll();
+            if (settingsForm.Visible)
+            {
+                // If helper window is not on top, this just focuses on the app again
+                // Pressing the ghelper button again will hide the app
+                if (!settingsForm.HasAnyFocus())
+                {
+                    settingsForm.ShowAll();
+                }
+                else
+                {
+                    settingsForm.HideAll();
+                }
+            }
             else
             {
 
@@ -279,5 +289,4 @@ namespace GHelper
 
 
     }
-
 }
--- a/app/Program.cs
+++ b/app/Program.cs
@@ -213,13 +213,13 @@ namespace GHelper
             SetAutoModes(true);
         }
 
-        public static void SettingsToggle(string action = "")
+        public static void SettingsToggle(string action = "", bool checkForFocus = false)
         {
             if (settingsForm.Visible)
             {
                 // If helper window is not on top, this just focuses on the app again
                 // Pressing the ghelper button again will hide the app
-                if (!settingsForm.HasAnyFocus())
+                if (checkForFocus && !settingsForm.HasAnyFocus())
                 {
                     settingsForm.ShowAll();
                 }
--- a/app/Settings.cs
+++ b/app/Settings.cs
@@ -547,6 +547,7 @@ namespace GHelper
             if (matrix == null || matrix.Text == "")
             {
                 matrix = new Matrix();
+                AddOwnedForm(matrix);
             }
 
             if (matrix.Visible)
@@ -615,6 +616,7 @@ namespace GHelper
             {
                 keyb = new Extra();
                 keyb.Show();
+                AddOwnedForm(keyb);
             }
             else
             {
@@ -643,6 +645,7 @@ namespace GHelper
             if (fans == null || fans.Text == "")
             {
                 fans = new Fans();
+                AddOwnedForm(fans);
             }
 
             if (fans.Visible)
@@ -839,6 +842,9 @@ namespace GHelper
             Application.Exit();
         }
 
+        /// <summary>
+        /// Closes all forms except the settings. Hides the settings
+        /// </summary>
         public void HideAll()
         {
             this.Hide();
@@ -849,14 +855,9 @@ namespace GHelper
 
         /// <summary>
         /// Brings all visible windows to the top, with settings being the focus
-        /// <br/>
-        /// Note: this will not respect previous focus i.e. will always focus settings
         /// </summary>
         public void ShowAll()
         {
-            if (fans != null && fans.Visible) fans.Activate();
-            if (keyb != null && keyb.Visible) keyb.Activate();
-            if (updates != null && updates.Visible) updates.Activate();
             this.Activate();
         }
 
--- a/app/Settings.cs
+++ b/app/Settings.cs
@@ -847,6 +847,27 @@ namespace GHelper
             if (updates != null && updates.Text != "") updates.Close();
         }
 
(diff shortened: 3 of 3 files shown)


nopeless:
@seerge for around `Settings.cs:613` where `#ButtonKeyboard_Click` is called, I am not sure why `keyb.Show()` has to be called. Removing that line does not show the keyboard panel, but fans and matrix doesn't have the `#Show()` but still shows fine. I think you should unify the behavior

AudreyAP:
This is awesome. Thank you!

seerge:
@nopeless great, thanks

seerge:
@nopeless I have unified forms behavior (as much as it was possible at least) and also added new logic from your PR to SettingsToggle called from tray icon click, as there it also won't harm ;)

AudreyAP:
This behaviour is not working for me on v128. Clicking the rog key does not being the window to the front. Sometimes, clicking the taskbar icon does. No relevant information in logs.

seerge:
@AudreyAP that's probably due to my tray-icon click tweaks :) i have re-uploaded https://github.com/seerge/g-helper/releases/tag/v0.128 with a small fix, try again

AudreyAP:
@seerge that fixed the tray icon click, but ROG button is still not working. Here's what I'm testing:

Open edge, full screen

Click M4, ghelper opens

Click edge

Click M4

Ghelper and fan tab does not come to front

Clicking the tray icon now works 100% exactly as expected though!

nopeless:
@AudreyAP what do you mean "clicking" rog key? its a physical key ok the keyboard

nopeless:
> @seerge that fixed the tray icon click, but ROG button is still not working. Here's what I'm testing:
> 
> Open edge, full screen
> 
> Click M4, ghelper opens
> 
> Click edge
> 
> Click M4
> 
> Ghelper and fan tab does not come to front. The icon at the bottom pulses like it is active, but the window doesn't appear. 
> 
> Clicking the tray icon now works 100% exactly as expected though!

nvm I just saw this. Weird, I'll check it out

AudreyAP:
Here's a video, just so we are on the same page:

https://github.com/seerge/g-helper/assets/130399040/48eb73bc-ec78-4e39-8f78-a2e6f2bfafa5

Here is my ghelper log: 

[log.txt](https://github.com/seerge/g-helper/files/12805340/log.txt)

Absolutely bone stock Windows 11 Pro for Workstations with modern nvidia drivers installed. No additional software installed apart from ghelper and nvidia drivers via nvcleanstall. Zero settings changes (just connected to wifi). It's literally bone stock windows 11.

nopeless:
@AudreyAP I confirmed the issue. https://github.com/nopeless/g-helper/actions/runs/6408603385

This run contains an unmodified executable from this PR. You can safely check if the executable was built without malware. Can you try it out?

AudreyAP:
@nopeless Exact same behavior as before. 

Deleted %appdata%/ghelper and deleted all previously downloaded copies of ghelper as well, just in case. Same issue.

nopeless:
@AudreyAP thanks for your time. Apparently this doesn't appear when its run from debug. I don't know the cause of this at the moment. I'll work on a fix

AudreyAP:
@nopeless No worries, thank you for working on it!

The taskbar behavior does seem to work as expected with the current mainline ghelper v0.128 prerelease, so it's a partial win :)

Seems like you found why I get the issue and you/serge don't, but please let me know if you need any further testing or anything.

nopeless:
It appears that I have stepped into a very infamous issue

https://www.gamedev.net/forums/topic.asp?topic_id=218484

I came up with a "fix" but here are the limitations

- Sometimes the window will be at the top but not have focus. Meaning that keyboard navigation will be sent to the active foreground window.

@AudreyAP Let me know what you think. The zip file will be available https://github.com/nopeless/g-helper/actions/runs/6409564963 once the run is completed.

There doesn't appear to be a clean working fix either. I have looked through countless posts and Windows documentation. This just doesn't appear to be a thing for some reason.

Anyone else is welcome to try and fix


The reason why this works perfectly in a debug environment is because the app is being debugged

The following is from Microsoft's official documentation

---

Remarks

The system restricts which processes can set the foreground window. A process can set the foreground window by calling SetForegroundWindow only if:

    All of the following conditions are true:
        The calling process belongs to a desktop application, not a UWP app or a Windows Store app designed for Windows 8 or 8.1.
        The foreground process has not disabled calls to SetForegroundWindow by a previous call to the [LockSetForegroundWindow](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-locksetforegroundwindow) function.
        The foreground lock time-out has expired (see [SPI_GETFOREGROUNDLOCKTIMEOUT in SystemParametersInfo](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfoa#SPI_GETFOREGROUNDLOCKTIMEOUT)).
        No menus are active.
    Additionally, at least one of the following conditions is true:
        The calling process is the foreground process.
        The calling process was started by the foreground process.
        There is currently no foreground window, and thus no foreground process.
        The calling process received the last input event.
        Either the foreground process or the calling process is being debugged.

AudreyAP:
> It appears that I have stepped into a very infamous issue

> April 09, 2004

Oh no lol. 

The fix seems to mostly work but there is one weird behavior - switching windows, then dismissing ghelper with m4, then opening ghelper with m4 sometimes results in the window being behind *only* the most recently activated window. 

I don't know if this is something you've already considered, but the existing "always on top" functionality seems to work perfectly. How about toggling that on momentarily? Don't know anything about how that is implemented, just spitballing :)

AudreyAP:
Maybe another issue - I think this also breaks "keep app window always on top" in extras menu?

It's crazy that there is no good way to do this in 2023... not even an app run as administrator can force its way to the top? So strange.

nopeless:
> Maybe another issue - I think this also breaks "keep app window always on top" in extras menu?
> 
> It's crazy that there is no good way to do this in 2023... not even an app run as administrator can force its way to the top? So strange. 

Oh, I should actually store whether topmost was toggled before toggling it.

You are right

nopeless:
> Maybe another issue - I think this also breaks "keep app window always on top" in extras menu?
> 
> It's crazy that there is no good way to do this in 2023... not even an app run as administrator can force its way to the top? So strange. 

There were some people quoting windows trying to prevent annoying popups. Who knows

AudreyAP:
https://www.reddit.com/r/csharp/comments/ledu64/setforegroundwindow_not_working_on_release_build/gmc7393/

Have you seen this? Kind of a horrifying hack, but seems like it should work?

nopeless:
With that being said, I'm tempted to try this

https://www.elevenforum.com/t/change-foregroundlocktimeout-to-open-app-window-in-front-in-windows-11.14363/

AudreyAP:
https://www.codeproject.com/Tips/76427/How-to-Bring-Window-to-Top-with-SetForegroundWindo

Some more hacks deep, deep in google search... Unbelievable that there is no setting to disable this, or allow certain apps, or override if admin. Thanks microsoft!

nopeless:
> https://www.reddit.com/r/csharp/comments/ledu64/setforegroundwindow_not_working_on_release_build/gmc7393/
> 
> Have you seen this? Kind of a horrifying hack, but seems like it should work?

yes, and its not really a good idea. There is a workaround that I liked

Its just minimizing and maximizing the app. This looks kinda funky but it always works. Would you want to try this?

AudreyAP:
@nopeless https://www.codeproject.com/Tips/76427/How-to-Bring-Window-to-Top-with-SetForegroundWindo

This seems to imply that you can literally just send an "alt" keypress while using setforegroundwindow... not at my laptop so can't test with one of the old builds but that seems perfect?

nopeless:
> @nopeless https://www.codeproject.com/Tips/76427/How-to-Bring-Window-to-Top-with-SetForegroundWindo
> 
> This seems to imply that you can literally just send an "alt" keypress while using setforegroundwindow... not at my laptop so can't test with one of the old builds but that seems perfect?

I am assuming that this will send an alt down to the program first. This isn't rly a problem for most programs as they take alt up (lifting up the alt key). I'm not a fan of the idea but I might try this
 @AudreyAP

nopeless:
@seerge I'm going to bite the bullet and try fixing this again. are you ok with that?

PR #1590, opened 2023-11-12
added window icon based on GPU mode

bourbonbourbon:
Since there is no window icon on the top left of the application, it looks a bit ugly when alt tabbing. I have also made it such that the window icon reflects the current GPU mode.

### DIFF
--- a/app/Settings.Designer.cs
+++ b/app/Settings.Designer.cs
@@ -1439,7 +1439,6 @@ namespace GHelper
             MinimumSize = new Size(822, 71);
             Name = "SettingsForm";
             Padding = new Padding(11);
-            ShowIcon = false;
             StartPosition = FormStartPosition.CenterScreen;
             Text = "G-Helper";
             panelMatrix.ResumeLayout(false);
--- a/app/Settings.cs
+++ b/app/Settings.cs
@@ -50,6 +50,22 @@ namespace GHelper
         {
 
             InitializeComponent();
+
+            int GPUMode = AppConfig.Get("gpu_mode");
+
+            switch (GPUMode)
+            {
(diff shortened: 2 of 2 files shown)


seerge:
@bourbonbourbon hello, thanks for PR

I like the idea of changing app icon with GPU mode (same as it does in tray) and also showing it in alt-tabbing. But the problem is that it doesn't look good on main windows itself, cause it pretty much says GG-Helper :)

image

Next to that, it's better to update icon in a same spot where tray icon is set (cause then it will update every time you change gpu mode and not only on startup) I have already added this change
https://github.com/seerge/g-helper/commit/940993e1d51e214bc104cbf2161caee248fd1100

bourbonbourbon:
> But the problem is that it doesn't look good on main windows itself, cause it pretty much says GG-Helper
Hmm you are right it does look weird. But now I am thinking that you should change the icon itself. Because this project now supports a wide range of laptop lineups. And probably change the name of the project as well? Hmm unless by "G" you mean gamer? Or you could do something like [Icon] - Helper in the title? Ahhh I can't think of anything viable tbh....

> Next to that, it's better to update icon in a same spot where tray icon is set (cause then it will update every time you change gpu mode and not only on startup)
Oh right I should have thought of this 😅

seerge:
@bourbonbourbon i prefer to keep main icon with the G letter in it, as it's part of the G-Helper name. 

And app is now widely known (50-60K active users), and popular on reddit, so name should stay.

seerge:
@bourbonbourbon i have added more lightweight icon (just a colored dot) to main window, it seem to fit better imho

Therefore closing this PR :) 

Screenshot 2023-11-13 102913

PR #1595, opened 2023-11-13
fix toggle touchpad

Oleg52:
After v0.131 update i noticed that FN+F10 doesn't toggle touchpad state at all, in the discussion [#1504](https://github.com/seerge/g-helper/discussions/1504) i meant to add delay before firing all key up events, now it works as it should, even with 50ms delay


seerge:
@Oleg52 ok, for me it keeps working in all situations anyway

PR #1635, opened 2023-11-22
Ambient 1 zone improve

DLdota:
if resolution is less than 512x288 after compression, then black color absorb other colors


seerge:
Oke!

PR #1639, opened 2023-11-23
fix: removed = from resolve conflict

Constrat:
With this I don't have any more commits + removed =.

This has been journey, sorry for all the mistakes :P


Constrat:
Ignore: already resolved with 36d6d6af48553de5a87b687161930b65a4a01330

PR #1651, opened 2023-11-25
Remove delay when main window opens

saneksem:
Calling `Startup.StartupCheck()` method may result to ~0.5-2 seconds delay before application window loads and becomes responsive. It depends on how many windows services are running and lag may happen on this line `var task = taskService.RootFolder.AllTasks.FirstOrDefault(t => t.Name == taskName);.` Sometimes it executes faster sometimes very slow.
I don't think we need to check task path on every window opening. If application was updated or moved then it should be enough to check task's exe path once on startup.

### DIFF
--- a/app/Program.cs
+++ b/app/Program.cs
@@ -122,6 +122,8 @@ namespace GHelper
                 SettingsToggle(action, false);
             }
 
+            Startup.StartupCheck();
+            
             Application.Run();
 
         }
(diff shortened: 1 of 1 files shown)


saneksem:
It doesn't happen every time. I attached GIF with visible delay. I think **it happens only when "Run on startup" is uncheked** and task service iterates ALL elements.
ghelper-lag

seerge:
@saneksem hello, I have moved whole action block out from ``SettingsToggle`` as it's actually not needed there indeed :) But thanks for spotting.

Closing this then

PR #1662, opened 2023-11-27
Added support for ROG Chakram Core (P511)

IceStormNG:
Adds support for the ROG Chakram Core (P511) mouse.


seerge:
@IceStormNG thanks!

PR #1700, opened 2023-12-07
Surpport Windows Style Toast

View12138:
GHelper Style Toast: [config.json] toast_mode:0
Windows Style Toast: [config.json] toast_mode:1
Muted_Toast


seerge:
@View12138 hello, thanks for contribution,

But for me your new toast window constantly shows some debug form below itself (even when it's invisible) :)

Also I see that some icons don't match (or don't change status) comparing to current implementation (like touchpad disabled)

View12138:
@seerge Thank you for your review!😊 The display logic for the new Toast has been revised, and new ICONs are now in use. However, the ICON display logic remains consistent with the previous version and is determined based on the ToastIcon enumeration.

PR #1762, opened 2023-12-19
Support for ASUS TuF M3 Gen II

IceStormNG:
ASUS released a Gen II version of the TuF M3 mouse. Same protocol, but different hardware identifiers.
This adds support for the Gen II variant.


seerge:
@IceStormNG by the way, do you have Discord or any other way to contact? I wanted to ask something :)

PR #1848, opened 2024-01-04
Support for Gladius II and Gladius II Origin

IceStormNG:
Adds support for Gladius II and Gladius II Origin (P502 and P504).


seerge:
@IceStormNG great :) big thanks as usual 👑

PR #1895, opened 2024-01-11
Update Program

Adam-S-Amir:
I love G-Helper but the built-in update button doesn't work, so this is a quick and dirty fix with bigger changes on the way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 I noticed that G-Helper's built in update button doesn't work too well so I made this simple batch file.
 
+I AM NOT AFFILIATED WITH THE G-HELPER PROJECT.
+
 # TODO
  - [ ] Make standalone application
  - [ ] Link to main app
--- a/README.md
+++ b/README.md
@@ -1,0 +1,7 @@
+# What is this?
+
+I noticed that G-Helper's built in update button doesn't work too well so I made this simple batch file.
+
+# TODO
+ - [ ] Make standalone application
+ - [ ] Link to main app
--- a/Updater.bat
+++ b/Updater.bat
@@ -1,0 +1,65 @@
+@echo off
+setlocal enabledelayedexpansion
+
+:: Check if running with administrator privileges
+NET SESSION >nul 2>&1
+if %errorlevel% neq 0 (
+    echo This script requires administrator privileges.
+    echo Please run the script as an administrator.
+    pause
+    exit /b 1
+)
+
+:: Your script code goes here
+
+echo Running with administrator privileges.
+
+REM Replace 'owner' and 'repo' with the GitHub owner and repository name
+set owner=seerge
+set repo=g-helper
+
+REM Fetch the latest release information
+for /f "tokens=*" %%i in ('curl -s "https://api.github.com/repos/%owner%/%repo%/releases/latest" ^| jq -r ".assets[0].browser_download_url"') do set download_url=%%i
+
+echo Download URL: !download_url!
+
+REM Specify the process name
+set "process_name=GHelper.exe"
+
+REM Kill the process
+taskkill /F /IM "%process_name%"
+
+powershell -Command "winget install -e --id jqlang.jq"
+
+echo Process %process_name% killed.
+
+:: Get the full path of the batch file's directory
+set "destination_folder=%~dp0"
+
+:: Loop through files in the batch file's directory
+for %%F in ("%destination_folder%\*") do (
+    set "file=%%~nxF"
+    if /I "%%~xF" neq ".bat" (
+        del "%%F"
+        echo Deleted: !file!
+    )
+)
+
+echo Files deleted, excluding batch files.
+
+REM Download the file
+curl -L -o "%destination_folder%\GHelper.zip" %download_url%
+
+REM Change the current directory to the destination folder
+cd /d "%destination_folder%"
+
+REM Unzip the file
+powershell -command "Expand-Archive -Path '.\GHelper.zip' -DestinationPath '.\'"
+
+REM Delete the zip file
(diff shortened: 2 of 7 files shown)


seerge:
@Adam-S-Amir hello, if you mean Auto Update prompt - it works :) what exactly doesn't work in your case?

I'm not a fan of creating separate executables, just to update

Adam-S-Amir:
@seerge Hey! Yes, we can download the zip file, but it won't self-extract and replace the current version of G-Helper with the latest build. This goes for the auto update prompt and the download button in the bottom left-hand corner. A separate executable would be useful as it would solve these issues, and linking a button to said executable would (in my opinion) make this application more user-friendly/dummy proof. This task would be taken care of by me if given the opportunity, though merging my own changes are up to you. Again, I love this app; I wanted to fix a niche issue I had.

seerge:
@Adam-S-Amir which version did you use to trigger auto update ? as in https://github.com/seerge/g-helper/releases/tag/v0.144 i have fixed an issue when it wouldn't extract if a folder with app contained special characters or spaces

So if you run 0.144 -> autoupdates should work (but there is no newer "final" release yet). It will trigger when i will mark new version as final :)

seerge:
@Adam-S-Amir i have marked 0.146 as "final" release, so 0.144 (with auto-update fix) should now trigger update dialog :) please check if it works for you

IceStormNG:
For me, the auto update did not work from 144 to 146. It closed and restarted, but started the old version again. I had to manually download the zip and replace GHelper

seerge:
@IceStormNG it starts powershell process with unzip script inside
https://github.com/seerge/g-helper/blob/df35a576dfa772b3f99e3c669dee1e827cbdeabe/app/AutoUpdate/AutoUpdateControl.cs#L122C17-L122C17

Can you check ``log.txt`` for for actual command and just try to run it manually in powershell ? Would it work ?

IceStormNG:
I guess I know what the issue is. My GHelper sits in "C:\Programs\GHelper". And when I run this manually, I get a "permission denied" when it tries to remove the old files.
Although, GHelper is running as admin, and should therefore not run into this issue. When I run the PS as admin, it does replace the files

seerge:
@IceStormNG ah! I can then launch powershell with same permissions as parent process runs then.

seerge:
@IceStormNG can you check if this build (with fake version 0.100) updates correctly in your conditions ?

[GHelper.zip](https://github.com/seerge/g-helper/files/13922782/GHelper.zip)

IceStormNG:
Nope. Does not work. It downloads the file, and also removes the archive afterwards, but it does not replace the GHelper files.

seerge:
@IceStormNG did you start it as admin ?

IceStormNG:
I run it through task scheduler, and the flag "run with highest permission" is ticket.

seerge:
@IceStormNG hm, I can't reproduce it now

I have placed test exe to ``C:\Program Files\GHelper`` (so system asked me for admin permissions even to create folder and copy it there)

Then if I run app as admin - it prompts for update - and auto updates
If I run it not as admin - it prompts for update and (expectedly) fails

I have already pushed this tweak to main https://github.com/seerge/g-helper/blob/527769042fc405354e2008fd25987727725a4c83/app/AutoUpdate/AutoUpdateControl.cs#L134

Mb you can tinker with that on your end, and try to figure out what prevents it from updating ?

IceStormNG:
https://stackoverflow.com/questions/2532769/how-to-start-a-process-as-administrator-mode-in-c-sharp

It looks like newer .Net neeeds UseShellExecute Set to true for the runas to work. I might try that later.

Kikimahe:
hi,

I had the same problem as @IceStormNG doing the updates : 

- Ghelper in this location: C:\Programs\GHelper
- Permission denied to save Zip file to this location.

I have tested the version above offered by Seerge.
Update works perfectly now.  👍

IceStormNG:
Did some more testing. The reason is, that GHelper did not exit quickly enough so it gets a "permission denied" as windows does not allow deleting running exectuables. I will try to find out how to solve that.

Kikimahe:
I don't think so because I tried this :
- Closing Ghelper
- Disabled task for Ghelper autostart
- Windows restart.

After restart, Windows still refuses to save Ghelper.zip to this location.

IceStormNG:
For me it is. I just tested it by using "Wait-Process -Name" instead of sleep. GHelper does not terminate. If I then exit it via taskmanager, the update runs through.

seerge:
@IceStormNG yeah, probably 1 second is not enough. Can you paste here full powershell command that you used instead ? or just do a PR with a tweak?

Thanks! :)

IceStormNG:
> @IceStormNG yeah, probably 1 second is not enough. Can you paste here full powershell command that you used instead ? or just do a PR with a tweak?

I opened a PR with the two changes. It now works fine for me.

seerge:
@IceStormNG merged :)

Kikimahe:
It's fix. 

Even though I am an administrator, I had to modify the "user" permissions of the `C:\Program Files\GHelper `folder by checking "total control".

And now the automatic update happens by itself. ! \O/

Capture d'écran 2024-01-13 121615

IceStormNG:
@Kikimahe  yeah, I also had to change permissions for that folder so that the regular user accounts can modifiy it. No big deal for me, although it should not be necessary.
For the admin to work, one has to set the "UseShellExecute = true" IIRC, but that makes the powershell window visible. And might scare some users xD

seerge:
Oke, thanks everyone for cooperation. I'm closing this PR as it's not needed anymore :)

PR #2004, opened 2024-01-30
LED brightness fix for TuF M3 Gen II

IceStormNG:
The TuF M3 Gen II uses 0-100 as brightness values instead of 0-4, which is used by the older Gen I variant.


seerge:
@IceStormNG amazing. Thanks

PR #2030, opened 2024-02-04
Mouse Profile Import/Export

IceStormNG:
This allows to save mouse profiles to a file on your computer and import them back later.
Might be helpful when messing with settings or with AC that likes to revert your mouse settings as soon as it detects your peripheral without even asking.

Writes GMP1 files (GHelper Mouse Profile Version 1).

Export/Import will operate on the currently selected profile. It does not export/overwrite other profiles besides the one currently selected.
All settings of the currently active profile are exported, except for which DPI profile is active, because not all mice support it.

Mouse profiles are mouse dependent. You cannot import the profile from another mouse that has different featureset.

image


seerge:
@IceStormNG ok, great. I assume it's indeed easier to have them separate from main app config.

IceStormNG:
> @IceStormNG ok, great. I assume it's indeed easier to have them separate from main app config.

Yes. I want to avoid what AC does. Having local storage and overwrite whatever is on the mouse all the time. 
It's more of like a backup feature. Especially if someone installs AC or connects the mouse to another PC with AC on, the mouse profile gets overwritten by AC immediately.
I guess most people don't need it, but it will be useful to some.

PR #2151, opened 2024-03-02
[Advanced Users only] CPU AutoTDP

IceStormNG:
This is an experimental implementation of AutoTDP for the CPU. Right now it works with Intel CPUs directly (requires the Winring0 drivers), and all ASUS laptops through the ASUS driver.

To work, RTSS (Riva Tuner Statistics Server) is required! If you don't have it, the feature will not be available to you. 
If you have MSI Afterburner installed, you already have RTSS installed. Afterburner and GHelper AutoTDP can run at the same time without issues (so you can have the overlay from Afterburner and let AutoTDP do the magic). 
Otherwise, you can install it from https://www.guru3d.com/download/rtss-rivatuner-statistics-server-download/

This feature is currently disabled on the ROG Ally... as there will be a special implementation for it at some point.

If your system meets all requirements, the AutoTDP button will show up below the GPU buttons.
image

### What is it?

First of all, this is NOT a framerate limiter. Instead, you should set a framelimiter for  the games you want to use it for in your GPU driver, either Nvidia Control Panel or AMD Radeon Software. 
Modern GPU drivers already have a AutoTDP like feature for their GPUs if you configure a framelimiter in the driver compared to using an external framelimiter or the one integrated into the game.

### So, what is it then?
It's an algorithm that constantly adjusts your CPU's power limits to be just as high as needed to run your game at the intended framerate. Especially on Intel CPUs, this can cut the CPU power usage down significantly, depending on CPU and game.
This will allow you to run the game at very similar performance but with less heat and less power draw. 
This is great if you want to game on battery, or want to reduce fan noise or are stuck with a USB-C charger that can not fully power your device.

### How to set it up

If you click on the AutoTDP button, a new windows will open which allows you to configure it.

image

The first drop downs allow you to do some general configuration, like what method to use to limit CPU power or what way to use to get the game's framerate. Right now, only RTSS can be used as source.

If enabled, it will still do nothing yet. You have to add the games that you want to use it for. It will not engage for any other application.

Click on the "Add Game" button and select the .exe file of your game that you want to use AutoTDP for. Keep in mind that some games have different exe, for example for dx12 or vulkan, or one is a launcher and the other the game (Fallout 4 does that). Make sure to use the one that is running your game.

Afterwards you can configure the settings for the game.

image

Select the same framerate, as the one that you configured your framerate limiter to. Btw: if you use this feature without a framelimiter, it will still work and will try to cut power to the CPU, trying to limit the game's framerate. However, this is not going to be smooth, so you really should set up a framelimiter.

The min and max TDP settings define the range that AutoTDP is allowed to adjust in. It will never exceed those values.

If you know how your CPU behaves, you can find those values rather quickly. Otherwise, you might have to experiment with it a bit. An overlay like Afterburner can be of help here to check how much power your CPU is consuming to get an idea of what is a useful range.

### Won't this cause stuttering?
Yes and no. There will be some slight framedrops when CPU load increases suddenly or when the algorithm tries to find the lowest stable power levels. Loading times can also be slower than usual. This heavily depends on the game though. Some games work without big issues, some will have performance issues. You can resolve those by increasing the min TDP by a few watts.
Again: if you play competitive games with super high framerates and need the best possible performance, do not configure AutoTDP for those games. You can still do if you wish and we won't stop you, but your right to complain is gone.

@seerge  if you want to, you can use this text as a "readme" to document the feature if you are going to merge it... and as a warning to users that should not touch it.

You might wanna check whether it works nicely on AMD devices, as I only have an Intel laptop for testing.


sirkuttin:
Super stoked for this! Excited for the ROG Ally impl.

Reylian2000:
Rog ally when please

RobinBeismann:
Hey @seerge,

any chance to get this merged once rebased and updated to the latest base version of g-helper?

seerge:
@RobinBeismann hello, eventually - may be. But currently my idea is to keep app as simple as possible. As it already has too much features as for me. 

There are couple of other niche branches (for example [Experimenal manual fan control](https://github.com/seerge/g-helper/discussions/2272)) that are separate as well :)

PR #2159, opened 2024-03-03
Mouse fixes

IceStormNG:
Fixes the DPI settings for the Strix Carry.
Adds support for the Strix Impact II Electro Punk, which is just a variant of the regular Impact II


seerge:
Thanks!

PR #2215, opened 2024-03-13
Add localization for some toast strings

emonq:


seerge:
@emonq hello, thanks for contribution. 

Problem is that translations are managed via [Crowdin ](https://crowdin.com/project/g-helper/) , and if you add them directly they screw up crowding readings :)

I was moving some new strings to translations as well, and included your suggestions in my own commit (with some minor adjustments).

So now, if you could just sign up on crowding, and translate them there - it would be perfect.

Thanks :)

emonq:
Thanks! I will update translations on crowdin.

PR #2305, opened 2024-03-27
Add Delay to Screenpad Actions

rollercoaster:
Add Delay to Screenpad Actions to make it smoother (similar to how asus apps apply it)

- Brightness changes are instant
- On/Off/Toggle changes are delayed
- delay can be set using "screenpad_delay" config, defaulting to 1.5 seconds

The main reason is to prevent jarring display connectivity changes that cause Windows 11 to rearrange windows, resolution, display positioning an brightness.


seerge:
@rollercoaster hello. Ok, i see that you really want a delay :) I will merge this then

seerge:
@rollercoaster change is live in 0.164 :)

PR #2644, opened 2024-05-28
Polling rate fix for Keris Wireless Aimpoint

etylix:
On newest firmware (April 2024), on packet 13, instead start from 0 for polling rate values, they bump it to 96. I have a check for someone that use the old firmware, tested on the new one but maybe wait for someone with old firmware check it.

### DIFF
--- a/app/Peripherals/Mouse/Models/KerisWirelssAimpoint.cs
+++ b/app/Peripherals/Mouse/Models/KerisWirelssAimpoint.cs
@@ -99,6 +99,18 @@
         {
             return true;
         }
+        // 3.00.06 - 4.00.01 or newer firmware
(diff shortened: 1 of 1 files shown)


seerge:
@etylix hello,

That's a good find! 

I don't have this mouse to actually test. Any chance you can downgrade firmware there just to verify that ? :)

Thanks.

etylix:
Sadly there is no way to downgrade (new firmware cut half of battery life lol), oh if you have any Asus mouse, you can change the main function to test it since most mouse use the old values.

seerge:
@etylix i see.

I have actually noticed that there is a fallback in a code, so i guess it's fine as is :)

Thanks.

PR #2651, opened 2024-05-29
Support for P504 Glaidus II PNK LTD

IceStormNG:
Support for P504 Glaidus II PNK LTD mouse. Dual color LED is not supported yet, as this is the only mouse so far that has this feature.


seerge:
@IceStormNG amazing, thanks a lot as usual!

PR #2685, opened 2024-06-04
Battery Percentage Pattern for Slash Lighting

lgarlati:
Added a new slash lighting pattern that reflects the current charge status of the machine. It respects the user's requested brightness level (dim, medium, bright) and sets the user's charge limit as 100% / full charge. 

Because the slash lighting is simply a string of bytes with no sense of memory of what set them, this pattern requires updating. TO accomplish this, a new timer was added that:
- only spawns when the user selects the battery pattern
- only refreshes the pattern every minute
- disposes of itself if somehow still running on a different pattern


seerge:
@lgarlati hello, thanks. I'm ok to merge it

PR #2795, opened 2024-06-30
Fixed DPI support for TuF M3 Gen II 

IceStormNG:
Fixed Support for TuF M3 Gen II. DPI configuration should now work as expected,

Note: TuF M3 Gen II (and maybe also other upcoming mice) enumerates the DPI profiles starting with 0 (first profile is 0x00), older mice start with 1 (first profile is 0x01).


seerge:
@IceStormNG thanks as usual :)

PR #2813, opened 2024-07-06
Tx gaming mini

IceStormNG:
Support for the TX Gaming Mini (P519). This is a Chinese variant of the TuF M4 Wireless


seerge:
@IceStormNG thanks :)

PR #2881, opened 2024-07-25
feat: bounce-cycle performance modes

Constrat:
This completely emulates how Armoury Crate handles the performance cycling through keybinds.

As the bounce mode is "dynamic", cycling "backwards" does not exist anymore.

Let me know what you think about it, if you would like to be the most similar to Armoury Crate or if you prefer the current logic

### DIFF
--- a/app/Input/InputDispatcher.cs
+++ b/app/Input/InputDispatcher.cs
@@ -373,7 +373,7 @@ namespace GHelper.Input
 
             if (e.Modifier == (ModifierKeys.Control | ModifierKeys.Shift | ModifierKeys.Alt))
             {
-                if (e.Key == keyProfile) modeControl.CyclePerformanceMode(true);
+                if (e.Key == keyProfile) modeControl.CyclePerformanceMode();
 
                 switch (e.Key)
                 {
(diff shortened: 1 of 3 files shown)


seerge:
@Constrat hello, that's nice, but I'm not sure if it makes sense to replace current way of cycling via modes backwards (with CTRL) and forwards with regular FN+F5. I have heard even few people complaining on reddit about AC's order of profiles and preferring the way G-Helper does that :)

Constrat:
Oh, interesting, welp. In case it's ever needed, the changes are here ;=)

seerge:
@Constrat i mean it's possible to put make it toggle-able via config, or via (another) hotkey .. but not to replace current method as it will cause lots of confusion otherwise i'm afraid :)

Constrat:
Yeah, it would be a pretty drastic change. I just made it out of spite as I used Armoury crate two days ago for testing, and I thought that G-helper didn't do that, so I put down some lines of code. To be fair I don't think I wasn't even going to use it, as I have a separate keyboard, other than the laptop. I guess I made it for "Completion" in case someone really liked it how AC handled it XD (hard to believe thinking about it)

PR #2962, opened 2024-08-11
Pro Art Hotkeys and visual modes support

seerge:


tristan-k:
How to invoke G-Helper on a Pro Art device like the PX13 if there is no dedicated M4/ROG key available? I tried to remap the Copilot (F23) key to G-Helper but sadly [PowerToys Keyboard Manager](https://github.com/microsoft/PowerToys/issues/3350) doesnt seem to have an option available to launch a specific app. Since the `Fn`-Key cant be mapped with PowerToys Keyboard Manager there is no way to launch G-Helper. Did I miss something?

seerge:
@tristan-k `CTRL+SHIFT+F12` should work on any keyboard, or a key that opens ProArt Hub (if there is one) too

tristan-k:
I must haved missed that the Fn + F12 Key is the ProArt key and as such I'm able to invoke G-Helper with it just fine. Sorry for the confusion.

PR #3032, opened 2024-08-30
Custom fan curves in tray icon menu

naixx:
This is WIP pr to add custom fan curves to the tray icon context menu
image

Currently, adding/removing fan curve doesn't update menu


naixx:
New/renamed/removed curves update tray icon context menu. 
PR is ready to be reviewed and merged

PR #3037, opened 2024-08-31
Add tooltip and message to specify seconds in keyboard light timeout

namar0x0309:
Hi,

I've had a hard time trying to figure out what the input boxes do as well as trying to understand what units the numbers represent. 
This is a small change to alleviate that.


seerge:
@namar0x0309 hello, thanks for pointing this out.

I have decided just to extend UI element itself, and added units directly inside it (as for some language translations label is already too long) :)

namar0x0309:
Awesome Thanks!

PR #3047, opened 2024-09-02
docs: add Japanese README

eltociear:
I created Japanese translated README.


seerge:
@eltociear どうもありがとう！ 

I have merged your translation. 
On a side note - application already supports Japanese language, but texts are not translated yet.

If you are enthusiastic enough - you can try to help here as well.

For translations I use service called Crowdin https://crowdin.com/project/g-helper/ja . You just need to join it and suggest translations there and then they will be automatically merged into main branch .

Thanks!

PR #3099, opened 2024-09-13
Added GA502IU model wih NoAura Devices

nicola02nb:
Added GA502IU model wih NoAura Devices because has an only static single color keyboard, with no animations; you can only dim it's brightness


seerge:
@nicola02nb ok, let it be :)

PR #3169, opened 2024-09-25
Properly disable OOBE mode

dantmnf:
Originated from https://github.com/torvalds/linux/commit/d6de45e3c6f3713d3825d3e2860c11d24e0f941f and also tested on Zenbook S 14 (UX5406SA).

Without setting `0x0005002F` to `1`, the keyboard backlight will keep breathing and uncontrollable.

Not sure how it will behave on other models though.

### DIFF
--- a/app/AsusACPI.cs
+++ b/app/AsusACPI.cs
@@ -114,7 +114,7 @@ public class AsusACPI
     public const int APU_MEM = 0x000600C1;
 
     public const int TUF_KB_BRIGHTNESS = 0x00050021;
-    public const int VIVO_KB_BRIGHTNESS = 0x0005002F;
+    public const int KBD_BACKLIGHT_OOBE = 0x0005002F;
 
     public const int TUF_KB = 0x00100056;
     public const int TUF_KB2 = 0x0010005a;
@@ -415,6 +415,14 @@ public class AsusACPI
 
         return BitConverter.ToInt32(status, 0) - 65536;
 
+    }
+
+    public bool DevicePresent(uint DeviceID) {
+        byte[] args = new byte[8];
+        BitConverter.GetBytes((uint)DeviceID).CopyTo(args, 0);
+        byte[] status = CallMethod(DSTS, args);
+
+        return (BitConverter.ToUInt32(status, 0) & 0x10000) == 0x10000;
     }
 
     public byte[] DeviceGetBuffer(uint DeviceID, uint Status = 0)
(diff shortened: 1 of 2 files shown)


seerge:
@dantmnf hello, thanks

Can you clarify a bit what this endpoint does ? cause based on other user's test it seem to be backlight brightness toggle (i might be wrong tho). What does OOBE stands for ?

dantmnf:
I think this should be a "feature" to indicate the system is starting by turning keyboard backlight on and off.

The term OOBE (out-of-box experience) is taken from Linux `asus-wmi` source, seems to be also reverse-engineered.

The two affected models AFAIK are on shining new platform (AMD Strix Point / Intel Lunar Lake), I'm not sure if Asus repurposed the endpoint in this generation.

seerge:
@dantmnf ok, then it could be that I have misunderstood sense of this in a first place.

So to summarize

Setting 0x0005002F to 1 - enables normal backlight operation and setting it to 0 - makes keyboard blink ?

dantmnf:
> setting it to `0` - makes keyboard blink ?
I haven't checked if this will make keyboard blink _again_ - obviously not a feature wanted for daily use, but it do have value `0` after system startup.

seerge:
@dantmnf but can you set it to 0 and check what effect it has ? Does keyboard start blinking ?

dantmnf:
Ok I will check it when the device comes back to me - maybe a few days later

dantmnf:
> @dantmnf but can you set it to 0 and check what effect it has ? Does keyboard start blinking ?

It has no effect. The keyboard won't start blinking again until a full shutdown and restart.

seerge:
@dantmnf ok, great. Then I have misinterpreted sense of this endpoint based on on feedback of other users.

Since OOBE resets on reboot, i assume it will also reset on hibernate / wakeup from it. So it makes sense to run it every time device changes power mode.

I have added OOBE command to TUF Keyboard Power method instead (responsible for backlight zones in different states and running every time power is changed)
https://github.com/seerge/g-helper/blob/740a34fa9c858171a0508404b4f2b211ff6c71e3/app/AsusACPI.cs#L833

https://github.com/seerge/g-helper/commit/534dc3d344ea47eab38fa52d8add6e58c61882f2

Can you check if this build works properly ?
[GHelper.zip](https://github.com/user-attachments/files/17188407/GHelper.zip)

dantmnf:
> i assume it will also reset on hibernate / wakeup from it.

You're right. I didn't think about the hibernate case.

> Can you check if this build works properly ?

It works well.

seerge:
@dantmnf ok! then i assume it's fine as is now? Thanks for spotting and clarifying!

memnon8711:
I have a new Vivobook S16 and have this issue when turning my laptop on.  Do I need to trigger something in GHelper to stop the rainbow puke or is this just supposed to work out of the box with GHelper?

Side note - thanks for adding my laptop to OLED list.

seerge:
@memnon8711 you don't need to do anything, as far as you use latest release - it will set this command as soon as app starts. You can verify it by checking if record mentioning `VIVO OOBE` exists in `%appdata%\GHelper\log.txt` after app start.

memnon8711:
I do see the line "VIVO OOBE = 1 : OK" in the log file.  I just notice the rainbow RGB comes one when I turn on the computer and I have to cycle through the brightness to turn it off.  Small nit.

seerge:
@memnon8711 can you check if this build behaves differently ?
[GHelper.zip](https://github.com/user-attachments/files/17490173/GHelper.zip)

memnon8711:
That worked.  The rainbow RGB was displaying when it booted up, but once Windows and GHelper loaded, it turned off the lighting and went back to the state I had it in previously.  Thanks for the quick response.

seerge:
@memnon8711 ok, great . I just swapped OOBE command and initial backlight setting that app sets

PR #3203, opened 2024-10-03
IsOLED: Add Zenbook S 14 (Intel), Zenbook S 16 (AMD)

dantmnf:
Also I have just spotted there are both OLED and non-OLED SKUs under the same model:

https://www.asus.com/laptops/for-home/zenbook/zenbook-14x-ux5401-11th-gen-intel/techspec/
https://www.asus.com/laptops/for-home/zenbook/zenbook-14x-oled-ux5401-11th-gen-intel/techspec/

Maybe we need some new approach to check OLED.

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -454,7 +454,7 @@ public static class AppConfig
 
     public static bool IsOLED()
     {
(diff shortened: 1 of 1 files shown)


seerge:
@dantmnf ok, thanks

PR #3315, opened 2024-10-27
Updated README.zh_CN.md, fixes title

pasical:
Some typography has been changed, and title jumps have been fixed


seerge:
@pasical hello, thanks for a contribution, but I don't really understand sense of it. You have made main download link very small and not prominent, why ?

pasical:
> @pasical hello, thanks for a contribution, but I don't really understand sense of it. You have made main download link very small and not prominent, why ?

Thanks for reviewing my PR. Well, in general, I should indeed be faithful to the original, but here the situation is slightly different. The main idea of this section should be the "Introduction," allowing users to understand the project before downloading. Typically, new users start with the "Installation Guide" rather than jumping directly to the download link, so I adjusted the heading levels to give higher priority to the guide and introductory content.

Since the Wiki interface is only in English, the Chinese version is part of the README, making the introductory section even more important. By organizing the headings this way, I aimed to create a more logical structure that guides users through the initial steps before downloading.

Of course, you can change it back if you feel it's really wrong.😊

PR #3358, opened 2024-11-10
Add support for CameraShutter

yinyue200:
Add support for CameraShutter on UX5401 and UX5406 to resolve the issue of the LED light not responding on the corresponding button.

See also: https://github.com/Kuchteq/asus-shutter-linux

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -698,10 +698,15 @@ public static class AppConfig
     {
         return ContainsModel("G834JYR") || ContainsModel("G834JZR") || Is("force_miniled");
     }
+    public static bool IsCameraShutter()
+    {
+        return ContainsModel("UX5401") || ContainsModel("UX5406");
+    }
 
     public static bool SaveDimming()
     {
         return Is("save_dimming");
     }
+
 
 }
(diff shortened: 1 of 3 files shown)


seerge:
@yinyue200 hello,

Interesting, thanks for a contribution and a good find. I was looking for an endpoint for camera by myself, but other owners of laptops that have this hotkey were not really helpful :)

Do I understand correctly that ``Program.acpi.DeviceGet(AsusACPI.CameraShutter)`` on the device always returns correct shutter status ? (i.e. 0 or 1)

Cause if so, just checking for a value being >=0 (i.e. if endpoint exists in general) would be enough instead of hardcoding models list under `AppConfig.IsCameraShutter()` :)

seerge:
@yinyue200 I have replaced `IsCameraShutter()` with just direct endpoint value check based on this assumption. 

As for you it seem to always return a value, and on the device that does not support it (like mine) endpoint doesn't exist at all. This is how most of the endpoints work on Asus devices and how G-Helper usually hides unsupported features from UI :)

seerge:
@yinyue200 yes, all endpoints that DEVS (`DeviceGet`) command returns contain a "presence" bit, that's why to get actual value I deduct it from result (i.e. deduct 65536 from int value)

https://github.com/seerge/g-helper/blob/c6ea0b0e3ea20b35ab54b4f9709e7ef59e4d5353/app/AsusACPI.cs#L411

And then, if result is < 0 it means endpoint doesn't exist at all.

Pro-tip: You can add `"debug": 1,` to `config.json` in the app, and then under `Extra` section you will see a zoom-glass button that runs a scan over all endpoints on your device. May be you will find something interesting there :)

PR #3360, opened 2024-11-10
Allow manual disabling of all functions related to the GPU

yinyue200:
I noticed that when the software starts on my device, it throws some exceptions that are caught and properly handled. Despite my device not having any dGPU, the AsusFan.GPU endpoint still exists, which causes an unnecessary GPU panel to appear on my device. This PR aims to minimize exceptions during the normal execution of the software to slightly improve performance. More importantly, it will allow users to manually hide the GPU panel and automatically hide these related panels on devices that are known to lack a dGPU.

### DIFF
--- a/app/AnimeMatrix/AniMatrixControl.cs
+++ b/app/AnimeMatrix/AniMatrixControl.cs
@@ -37,7 +37,7 @@ namespace GHelper.AnimeMatrix
         {
             settings = settingsForm;
             if (!AppConfig.IsSlash() && !AppConfig.IsAnimeMatrix()) return;
-            
+
             try
             {
                 if (AppConfig.IsSlash())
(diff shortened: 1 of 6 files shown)


seerge:
@yinyue200 hello,

For Matrix checking, i kind of agree. But I was thinking about just setting a list of models that do have matrix (as it's mainly G14, M16 and few others) , instead of opposite.

But for AmdGpuControl - it's needed for iGPU readings as well :) (and this is lots of devices with AMD CPUs and Rog Ally)

seerge:
@yinyue200 ok, even tho I think it's a bit an overkill, I have added an option to skip nvidia / amd driver checks all together, and also hardcoded your model automatically. 

Mainly to hide GPU section in the UI. But keep in mind, that now you won't be able to see second fan speed that your device seem to have ?

Also your `` if (!Adl2.Load() || AppConfig.NoGpu())`` would still check for AMD driver most likely, as ADL2 comes first in a condition :) So I have swapped them

PR #3408, opened 2024-11-25
Added estimated remaining time on battery

The-Faulty:
Estimated time remaining is shown when hovering over tray icon or when the discharge/charge label is hovered over. Shows both time left on battery and time left to charge battery.


lelecorj20:
Nossa, ótima ideia, seria legal, também, poder substituir o ícone do app pela porcentagem da bateria ou taxa de recarga/descarga.

PR #3511, opened 2024-12-23
Max performance

seerge:
Added option to set High Performance Power Plan instead of Balanced under `Fans + Power` for CPU intensive tasks.

### DIFF
--- a/app/Mode/ModeControl.cs
+++ b/app/Mode/ModeControl.cs
@@ -115,19 +115,9 @@ namespace GHelper.Mode
 
             if (!AppConfig.Is("skip_powermode"))
             {
-                var powerMode = AppConfig.GetModeString("powermode");
-                if (!PowerNative.IsHighPerformance(powerMode))
-                {
-                    // Power plan from config or defaulting to balanced
-                    if (AppConfig.GetModeString("scheme") is not null)
-                        PowerNative.SetPowerPlan(AppConfig.GetModeString("scheme"));
-                    else
-                        PowerNative.SetBalancedPowerPlan();
-                }
-
                 // Windows power mode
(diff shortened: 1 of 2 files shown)


isaLuckett:
is it finally possible to make each mode have GUI option for custom Power Plan ?
Armoury Crate is hiding/deleting Turbo and Silent when its not used.

i have been using Custom Power Plan to prevent overheating and limit my processor state to 50% in Silent mode
and my processor is 32°C with the Fan at 0 rpm, it actually very useful when you know how to configure power plan.

EminGT:
image
What's the difference between "Best Performance" and "High Performance Plan" that just got added?

seerge:
- Best Performance is default Balanced Power plan (in control panel) + Best Performance power mode (in windows)
- High Performance - is standalone "High Performance plan" (in control panel) :) 

It can give a tiny-bit more performance for a CPU intensive tasks at cost of higher power consumption of course.

elfi-ox:
Do you have an Idea of why my Labtop would make creaking noises when set into that mode ?

seerge:
@elfi-ox what do you mean under "creaking noises" ?

elfi-ox:
it's a very subtle, faint sound, audible by sticking my ear precisely where the CPU is. It's like a shortcut sound that dosen't transmit well into the video ( but it's audible I swear, the volume just has to be turned up )

https://github.com/user-attachments/assets/e8333a47-e61f-4b2d-92cd-5acc53fbe8e9

seerge:
@elfi-ox most probably that's a coil whine from a bit higher power consumption :)

elfi-ox:
I see that make sense, thanks for the insight

PR #3566, opened 2025-01-07
Added a useful batch script that updates GHelper for you.

osomat123:
This pull request introduces a batch script to streamline the process of updating GHelper.
The script performs the following steps:

1. **Requests Admin Privileges:** Required to save the GHelper binaries in C:\Program Files\GHelper.
2. **Terminates the Current GHelper Process:** Ensures no conflicts during the update.
3. **Deletes Existing GHelper Binaries:** Cleans up outdated files from C:\Program Files\GHelper.
4. **Downloads and Extracts the Latest Version:** Retrieves GHelper.zip from GitHub, unzips it, and saves the binaries in the target directory.
5. **Cleans Up Temporary Files:** Deletes the downloaded GHelper.zip to avoid clutter.
6. **Restarts GHelper:** Launches the updated version of the application.

I have been using this script to update GHelper consistently for the past 5-6 months, and it has worked reliably in every instance.

### DIFF
--- a/updateGHelper.bat
+++ b/updateGHelper.bat
@@ -1,0 +1,31 @@
+:: Check for administrative privileges
+net session >nul 2>&1
+if %errorlevel% neq 0 (
+    echo Requesting administrative privileges...
+    powershell -Command "Start-Process '%~f0' -Verb RunAs"
+    exit /b
+)
+
+set G_HELPER_PATH="C:\Program Files\GHelper"
+set URL=https://github.com/seerge/g-helper/releases/latest/download/GHelper.zip
+
+:: Kill current running instance
+taskkill /IM GHelper.exe /F
+
+:: Go to GHelper Directory
+cd /D %G_HELPER_PATH%
+
+:: Delete old exe
+del GHelper.exe
+
+:: Download and unzip new version
+curl -L %URL% --output GHelper.zip
+powershell -Command "Expand-Archive .\GHelper.zip"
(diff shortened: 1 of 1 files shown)


seerge:
@osomat123 hello, thanks for your effort, but G-Helper already has a built in auto-update system.

Every time the new "final" release is published, app would offer you automatic update.
https://github.com/seerge/g-helper/blob/bdfef9fbdb89f557c01ee39bed752a1afc5b8258/app/AutoUpdate/AutoUpdateControl.cs#L48

Thanks

seerge:
@osomat123 what does it say in app log in that moment ? I think it would be helpful if you can reproduce it - and post a new issue with the log :)

Do you keep GHelper.exe in some folder that require admin permissions?

osomat123:
@seerge - I keep GHelper.exe in C:\Program Files\GHelper.
I think admin permissions are needed here. Is this why the auto update doesn't work?
Would running GHelper in Admin mode help here?

I'll try to get a hold of the logs the new time I see the auto update window pop up.

ghost-of-cerberus:
@seerge - I have the same issue but I'm certain it's not permissions related as it's in my user profile directory; however, the path has special characters which I assume is the issue.

Posting a new issue with log now.

seerge:
@ghost-of-cerberus could be due to a special characters in a path, I have answered in your [issue](https://github.com/seerge/g-helper/issues/3568)

seerge:
@osomat123 in your case running it as admin permissions might help, can you try it using an older version?

PR #3635, opened 2025-01-19
Add EPP control functionality and UI integration

A7GoD:
This change adds EPP slider to control panel for ROG ally.
Under Fans+Power section.

image

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -436,7 +436,7 @@ public static class AppConfig
     {
         return ContainsModel("GA401") || ContainsModel("GA402") || ContainsModel("GU604V");
     }
-    
+
     public static bool IsSlash()
     {
         return ContainsModel("GA403") || ContainsModel("GU605") || ContainsModel("GA605");
@@ -723,5 +723,9 @@ public static class AppConfig
         return Is("auto_status_led");
     }
 
+    public static void SetEPP(int value)
+    {
+        PowerNative.SetEPP(value);
+    }
 
 }
--- a/app/Fans.Designer.cs
+++ b/app/Fans.Designer.cs
@@ -60,6 +60,7 @@ namespace GHelper
             checkApplyFans = new RCheckBox();
             buttonReset = new RButton();
             comboBoost = new RComboBox();
+            sliderEPP = new TrackBar();
             panelSliders = new Panel();
             panelAdvanced = new Panel();
             panelAdvancedAlways = new Panel();
@@ -113,6 +114,15 @@ namespace GHelper
             panelBoostTitle = new Panel();
             pictureBoost = new PictureBox();
             labelBoost = new Label();
+            if (AppConfig.IsAlly())
+            {
+                panelEPP = new Panel();
+                panelEPPTitle = new Panel();
+                labelEPP = new Label();
+                labelEPPSliderLeft = new Label();
+                labelEPPSliderRight = new Label();
+                pictureEPP = new PictureBox();
+            }
             panelPowerMode = new Panel();
             comboPowerMode = new RComboBox();
             panelPowerModeTItle = new Panel();
@@ -189,6 +199,17 @@ namespace GHelper
             ((System.ComponentModel.ISupportInitialize)pictureBoxCPU).BeginInit();
             panelBoost.SuspendLayout();
             panelBoostTitle.SuspendLayout();
+
+            if (AppConfig.IsAlly())
+            {
+                ((System.ComponentModel.ISupportInitialize)pictureEPP).BeginInit();
+                panelEPP.SuspendLayout();
+                panelEPPTitle.SuspendLayout();
+                labelEPP.SuspendLayout();
+                labelEPPSliderLeft.SuspendLayout();
+                labelEPPSliderRight.SuspendLayout();
+            }
+
             ((System.ComponentModel.ISupportInitialize)pictureBoost).BeginInit();
             panelPowerMode.SuspendLayout();
             panelPowerModeTItle.SuspendLayout();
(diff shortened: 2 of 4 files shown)


seerge:
@A7GoD hello,

That's an interesting find. But does this option affect actual processor behavior (especially AMD) cause form what I googled it seem to be specific for Intel?

Also how does it corelate with Power Modes (Best Power Efficiency / Balanced / Best Performance) in windows ?

A7GoD:
Hi @seerge, 
Glad you replied. From what i found when looking online there isn't mention of epp's correlation with manufacturer specific behaviour. AMD and Intel both seem to have different names for things having epp. What I think points closest to epp relating with epp seems to be this article: https://docs.kernel.org/admin-guide/pm/amd-pstate.html

Notice that the value ranges from

> Energy Performance Preference (EPP) (RW)¶
> This attribute provides a hint to the hardware if software wants to bias toward performance (0x0) or energy efficiency (0xff).

This is quite similar to windows' implementation:

https://learn.microsoft.com/en-us/windows-hardware/customize/power-settings/options-for-perf-state-engine-perfenergypreference

> When set to 100, the energy performance preference register is set to 255 to favor energy savings. When set to an intermediate value, the energy performance preference register is programmed to the value: (setting * 255) / 100.

Furthermore i have noticed that when windows mode is set to "silent" in ally, it by default seems to set epp to 60.
I think that the power modes you mentioned don't seem to change epp settings.

I would like to hear your views regarding it, hope I can contribute something to this beautiful project :)
Cheers

seerge:
@A7GoD ok,

I did some benchmarks to see if this setting is actually affects something.

I did run Cyberpunk benchmark on same graphic settings and same power limits (15W and 30W) using my Ally Z1 and set EPP slider to edge states - 0% and to 100% respectively. But benchmark results were pretty much identical, so it seems either AMD CPU ignores this setting completely or windows game mode overrides it and does it's job in balancing power already.

I also did same benchmark on my AMD G14 using iGPU only and iGPU + dGPU even. And again - same identical results.

Did you do any tests by yourself ? Do you see any real-life impact from this setting ?

A7GoD:
> @A7GoD ok,
> 
> I did some benchmarks to see if this setting is actually affects something.
> 
> I did run Cyberpunk benchmark on same graphic settings and same power limits (15W and 30W) using my Ally Z1 and set EPP slider to edge states - 0% and to 100% respectively. But benchmark results were pretty much identical, so it seems either AMD CPU ignores this setting completely or windows game mode overrides it and does it's job in balancing power already.
> 
> I also did same benchmark on my AMD G14 using iGPU only and iGPU + dGPU even. And again - same identical results.
> 
> Did you do any tests by yourself ? Do you see any real-life impact from this setting ?

Hi @seerge , I tried to test these myself, and couldn't notice changes, it only happened once for me that gpu clocks were as high as 2700 when epp was set to 100 but when i switched back to 0 it went to normal behaviour not changing again, i will need to dig into the issue more, seems like something is making the system ignore epp settings.
Thanks for testing it from your side as well. I have also noticed that auto controller mode is not working for me, so will try to fix that as well.

seerge:
@A7GoD CPU clocks in idle state already scale quite well when you change windows Power Modes (Power Efficiency would give you same 2.7Ghz)

A7GoD:
> @A7GoD CPU clocks in idle state already scale quite well when you change windows Power Modes (Power Efficiency would give you same 2.7Ghz)

Oh, I meant gpu clocks were 2.7ghz
But which checking the script made by ciphray.bat I found he is setting value for PERFAUTONOMOUSWINDOW as well and disabling gamemode. Seems like there is something that can be done to switch gpu priority, i am trying to grasp what that is. Hopefully I might be able to figure it out.

A7GoD:
> > @A7GoD CPU clocks in idle state already scale quite well when you change windows Power Modes (Power Efficiency would give you same 2.7Ghz)
> 
> Oh, I meant gpu clocks were 2.7ghz
> But which checking the script made by ciphray.bat I found he is setting value for PERFAUTONOMOUSWINDOW as well and disabling gamemode. Seems like there is something that can be done to switch gpu priority, i am trying to grasp what that is. Hopefully I might be able to figure it out.


@seerge  i have also noticed that outside any game, when the load is less on both CPU and GPU the slider works as intended as you can see the cpu clocks in both cases:





When epp is 100% the cpu clocks are considerably lower.

Also, that only works when power plan is applied using ciphray.bat's power profile, and not when set via Asus. So my guess is that epp is not that much effective and cpu starts ignoring the values as soon as it sees higher load. For now am not sure how can this be fixed, as asus and windows seem to start ignoring the epp as soon as they can, which is a bummer. It seems on linux the behaviour is much more consistent with the settings

seerge:
@A7GoD yeah, they are lower - but just changing Power Mode lowers clocks even more. Also keep in mind that on your screenshots you use a `High Performance` **Power Plan** (not default `Balanced` Power Plan even). It's supposed to keep clocks to the roof - cause it was the whole purpose of adding it (for CPU-heavy tasks).

My main observation was that I didn't notice any real-life changes from EPP sliders in apps / games

A7GoD:
> @A7GoD yeah, they are lower - but just changing Power Mode lowers clocks even more. Also keep in mind that on your screenshots you use a `High Performance` **Power Plan** (not default `Balanced` Power Plan even). It's supposed to keep clocks to the roof - cause it was the whole purpose of adding it (for CPU-heavy tasks).
> 
> My main observation was that I didn't notice any real-life changes from EPP sliders in apps / games.
> 
> If you want to watch videos / save battery - people can use Best Power Efficiency mode, if you want to game - use any other power mode.

Yes, you're correct, I didn't notice any change either, idk how people in forums and discord servers setup their ally, they seem to mention epp working.

PR #3814, opened 2025-03-09
Some fixes to spanish in the app

MemoSainz:
I just added some orthographic corrections in the language. Hope they will be useful.


seerge:
@MemoSainz hello, thanks for a contribution :) 

Currently all strings in the app are handled by Crowdin https://crowdin.com/project/g-helper . So I don't update them in the code directly (as then it can de-sync with translation platform). Can you please make your edits there directly ?

For your other PR with `Enable GPU on shutdown` - it's also in Crowdin. I agree that text could be more detailed, but main problem that it does not fit into a single line, especially in some non-english translations.

I might add a tooltip (same as it's done for a Clamshell mode) tho, explaining what it does.

PR #4071, opened 2025-05-14
Added Arabic localization (Strings.ar.resx)

asim1801:
Hi @seerge 

This PR introduces Arabic language support to G-Helper.

I have translated the existing English strings from `Strings.resx` into Arabic and included them in the new `Strings.ar.resx` file. 

However, during the translation process, I noticed several UI text elements that appear to be hardcoded directly in the source files rather than being sourced from the resource files. 

Here are some examples I spotted:

*   **Main UI Elements:**
    *   "Slash Lighting" (and its modes "Bounce, Slash, Loading, Bitstream and whatnot", "Interval Off", and interval values like "Interval Xs") appears in `app/Settings.cs`.
    *   "FN-Lock" (and its on/off states for button text and toast notifications) seems to be in `app/Settings.cs` and `app/Input/InputDispatcher.cs`.

*   **Extra Settings UI (`app/Extra.cs` and `app/Extra.Designer.cs`):**
    *   Section title "CPU Cores Configuration" and associated labels/dropdown text like "Pcores", "Ecores", and the "Apply" button.
    *   The term "action" next to the M-key/FN-key binding dropdowns.
    *   Time unit suffixes like "(sec)" and "(min)" for input fields.
  

Screenshot 2025-05-14 211819

### DIFF
--- a/app/Properties/Strings.ar.resx
+++ b/app/Properties/Strings.ar.resx
@@ -1,0 +1,821 @@
+<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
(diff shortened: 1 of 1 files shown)


seerge:
@asim1801 hello, that's great. Thanks for translations. Merged.

Actually they are managed by https://crowdin.com/project/g-helper/ system. So from now on, just edit them there. As otherwise crowdin-generated PRs conflict with manual ones.

Thanks!

PR #4074, opened 2025-05-15
Fixed the alignment of '+' button and mode dropdown in Fan Curves UI

asim1801:
Changed the size and location properties of 'buttonAdd' and 'comboModes' in 'app/Fans.Designer.cs' to ensure they are vertically aligned. A fix nobody asked for, but it was bothering me 😂. 

Before the change:
old

After the change:
new

### DIFF
--- a/app/Fans.Designer.cs
+++ b/app/Fans.Designer.cs
@@ -31,14 +31,14 @@ namespace GHelper
         /// </summary>
         private void InitializeComponent()
         {
-            ChartArea chartArea1 = new ChartArea();
-            Title title1 = new Title();
-            ChartArea chartArea2 = new ChartArea();
-            Title title2 = new Title();
-            ChartArea chartArea3 = new ChartArea();
-            Title title3 = new Title();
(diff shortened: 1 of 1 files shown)


seerge:
@asim1801 hello, with this changes ` + ` button shifts somewhere in the middle of the screen for me (200% scaling in Win) 

But since you have mentioned,[ I have re-worked that part of UI completely](https://github.com/seerge/g-helper/commit/00dfe01abbd146da8a347016589947d2a34b4b83), by arranging buttons inside mini-table.

So this build  should do a thing.

Please keep in mind that winforms used in the app, are generally absolutely terrible in terms of scaling in windows. I design everything around 200%. But if you use something very different from it (like 100%), at least vertical positioning of the buttons can shift a bit in comparison to dropdown.

[GHelper.zip](https://github.com/user-attachments/files/20242520/GHelper.zip)

asim1801:
Well, yeah, just tested it out, it works like a charm

image

PR #4080, opened 2025-05-16
Added localization for Slash Modes and Interval Settings

asim1801:
I updated app/AnimeMatrix/SlashDevice.cs so all 17 slash mode names (like "Bounce", "Slash", "Loading", "Bit Stream", etc.) now use Properties.Strings instead of hardcoded values. Also did the same thing to Settings.cs to build the interval dropdown using localized strings. 

Tested it with Arabic strings, it works fine: 
Screenshot 2025-05-16 155843

And that is how it looks in English after the change aswell: 
image


seerge:
@asim1801 ok, thanks. Even tho I'm not a big fan of translating things that are most likely not translated in AC :)

PR #4096, opened 2025-05-19
Add Overvolt setting for Power Users 

benlf1:
This is a very simple change that allows for CPU Overvolting in the config file by adding a "max_uv" variable.

I have an odd GA401QM that freezes/BSODs on idle when using the "Best Power Efficiency" power mode without a 15mV overvolt.


seerge:
@benlf1 ok, it's fine to make this param configurable :)

PR #4428, opened 2025-08-08
Extended battery limits

rorycondict:
Some laptops like my G16 2025 don't support charging limits outside of 60 to 80 percent

This PR adds a workaround to allow limits from 80 to 100 - the limit is set to 100 until it detects the charge equals or exceeds the user's setting, then its reduced to 60 which actually has the effect of keeping the battery at the current level

It is a bit hacky though, requires constantly monitoring the windows battery level and setting the limit accordingly, I've set the interval at 5 seconds per poll at least so I don't think resource usage should be a concern

Also unfortunately below 60 is still not supported as 60 is the lowest limit afaik...?

Let me know what you think and if you see any issues with it, new to this codebase, could probably use testing on the other machines with these limits too

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 ##
 ## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
 
+# JetBrains Rider
+*.idea
+
 # User-specific files
 *.rsuser
 *.suo
--- a/app/Battery/BatteryControl.cs
+++ b/app/Battery/BatteryControl.cs
@@ -1,4 +1,5 @@
 ﻿using System.Diagnostics;
+using System.Management;
 
 namespace GHelper.Battery
 {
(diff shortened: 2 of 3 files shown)


seerge:
@forthfora hello, thanks. I was also thinking about such approach. But I'm not sure if it's worth to try to outsmart hardware (and firmware) in a first place. Cause Asus restricted it to 80% for a reason most likely.

Also some devices, for example my older G14 2022 can discharge when lower limit is set on the fly. So such approach will simply make battery constantly charge up and down.

rorycondict:
Yeah makes sense, I suspected it was disabled for a reason, just would've been nice to be able to set a 90% or 95% limit as a balance

Maybe GHelper can support plugins someday for more experimental tweaks like this? Not sure the best way to implement that, maybe some sort of hooking system. Just something I thought about though, having my own fork works fine for me

PR #4429, opened 2025-08-08
Battery Time Prediction

forthfora:
Adds an option to estimate the battery time, available in Extras and disabled by default

(I would've used windows predictor but it seems like it was removed in 11...? I can't seem to enable it anyway, even through registry)

Just a simple prediction based on current discharge rate and charge level

image

image

Appears when hovering over the tray icon as well

Let me know if the huge diff is fine too, I've never messed with winforms before, so I wasn't sure if these UI changes are normal or easy to deal with when collaborating?

### DIFF
--- a/app/Battery/BatteryControl.cs
+++ b/app/Battery/BatteryControl.cs
@@ -109,12 +109,12 @@ namespace GHelper.Battery
             {
                 var estimateTimeToEmpty = TimeSpan.FromHours((double)(HardwareControl.batteryCapacity / Math.Abs((decimal)HardwareControl.batteryRate)));
 
-                return Strings.EstimatedToEmpty + ": " + EstimatedTimeToString(estimateTimeToEmpty);
+                return Strings.EstimatedBatteryRemaining + ": " + EstimatedTimeToString(estimateTimeToEmpty);
             }
 
             var estimatedTimeToLimit = TimeSpan.FromHours((double)((HardwareControl.batteryCapacity - (HardwareControl.chargeCapacity / 1000)) / HardwareControl.batteryRate));
 
-            return Strings.EstimatedToLimit + ": " + EstimatedTimeToString(estimatedTimeToLimit);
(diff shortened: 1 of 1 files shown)


seerge:
@forthfora hello, thanks but windows already has quite good (from my opinion) battery life predictor ? You can see it when hover mouse over battery icon in tray, or open settings. It also doesn't just use current (instant) discharge rate but rather a pattern of last X minutes (i don't know how much exactly).

rorycondict:
> @forthfora hello, thanks but windows already has quite good (from my opinion) battery life predictor ? You can see it when hover mouse over battery icon in tray, or open settings. It also doesn't just use current (instant) discharge rate but rather a pattern of last X minutes (i don't know how much exactly).

I can't get it to show up for some reason, already tried a registry tweak, it's just not there anymore (it was there on windows 10 on my last laptop, never appeared on this one)

Would happily use it if I could get it to show up

seerge:
@forthfora 
Screenshot 2025-08-08 190853
Screenshot 2025-08-08 190906
I'm talking about this one. I think it's pretty much in any windows? Did you try running troubleshooter / restoring damaged files in windows ? may be something is broken specifically for you : )

rorycondict:
Yeah I'll try troubleshooting again, here's how it looks for me lol
image

image

Anyways feel free to close the PR if you don't think it'll be useful, I'll keep it for personal use

seerge:
@forthfora try to completely uninstall battery driver from device manager (advice from ChatGPT)

As for the actual estimation. My point is that it will fluctuate too much if calculated instantly. I.e. my laptop varies between 5w and 7.5w discharge rate on battery, in reality it would result into like 50% difference in battery life difference :) And calculating something more sophisticated - is much more difficult + windows already kind of does it (even tho with lots of hit and miss too)

rorycondict:
alright thanks for the tip, I'll go ahead and close this for now

PR #4487, opened 2025-08-29
add Zenbook 14 intel (225H) to OLED devices

himazawa:


seerge:
@himazawa hello, your device should be already in the list as it would match `UX34` pattern in `ContainsModel("UX34")`

himazawa:
Oh sorry you are right, I will close the issue

PR #4508, opened 2025-09-02
feat: add unified fan control mode with CPU/GPU priority options

pashza:
The heat sinks are connected and when the GPU fan is not spinning the heat dissipation is worse, than when both fans move the same amount of heat.
especially on older models of the m16 since the intel cpu over heat rapidly.

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -757,6 +757,16 @@ public static class AppConfig
     {
         return ContainsModel("G834JYR") || ContainsModel("G834JZR") || ContainsModel("G634JZR") || ContainsModel("G835LW") || Is("force_miniled");
     }
+    public static int GetFanUnifiedMode()
+    {
+        return Get("fan_unified_mode", 0);
+    }
+
+    public static void SetFanUnifiedMode(int mode)
+    {
+        Set("fan_unified_mode", mode);
+    }
+
     public static bool SaveDimming()
     {
         return Is("save_dimming");
@@ -766,6 +776,4 @@ public static class AppConfig
     {
         return Is("auto_status_led");
     }
-
-
 }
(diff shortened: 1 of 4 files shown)


seerge:
@pashza hello, thanks for the effort. 

But there is [Experimental build with a special manual fan control](https://github.com/seerge/g-helper/discussions/2272) that already allows you to achieve whatever you want  :)

If app would just spam custom fan curves via standard armoury endpoints - it won't allow fan control. As it's still up to firmware how to handle fans in this case. And on most (especially new) laptops  due to a very high hysteresis reaction time will be like 1 minute even if you set a flat 100% fan curve.

pashza:
Hu Seerge,
What i added basically ties the gpu fan to the cpu fan curve.The cpu load is very small and not noticeable at all from testing, the point is that both fans spin at the same speed.It works instantly, i didnt see big difference in speeds.
I really think it works better than manual speed control as it is depending on cpu temperature which i found to be more important for computer performance than gpu temperature - at least for my requirements.
Check it out, ill check the experimental build.  
On Tue, Sep 2, 2025 at 12:55, Serge ***@***.***&gt; wrote:    seerge left a comment  (seerge/g-helper#4508)
 @pashza hello, thanks for the effort. But there is Experimental build with a special manual fan control that already allows you to achieve whatever you want :) If app would just spam custom fan curves via standard armoury endpoints - it won't allow fan control. As it's still up to firmware how to handle fans in this case. And on most (especially new) laptops due to a very high hysteresis reaction time will be like 1 minute even if you set a flat 100% fan curve. —

seerge:
@pashza experimental build can do same thing,

If you want both fans to follow max of CPU & GPU temps just set (as mentioned in description) :)

```
"fan_shift" : 100,
```

pashza:
i tested the experimental version - it so much more complicated than what i implemented. 
also my version doesnt require any changes to drivers or installing anything additional.
it follows either GPU or CPU temp for both fans. regardless of drivers.
it works instantly pretty much.
another thing is that its settable from the interface in an easy way.
with default being each fan follows its own.

PR #4524, opened 2025-09-05
NVidia services Stop / Start option

seerge:
Added option under `Extra` to `Stop Nvidia Services` when GPU is disabled and Start them back when GPU is enabled. 
It seem to give some battery saving advantages (4-4.7w discharge in idle state)

> [!NOTE]
> App needs to run with admin permissions for the feature to work

Screenshot 2025-09-05 154328

# [GHelper.zip](https://github.com/seerge/g-helper/releases/download/v0.231/GHelper.zip)

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -654,10 +654,6 @@ public static class AppConfig
     {
         return Is("nv_platform");
     }
-    public static bool IsNVPlatformExists()
-    {
-        return Exists("nv_platform");
-    }
 
     public static bool IsForceSetGPUMode()
     {
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -668,6 +668,10 @@ public static class AppConfig
     {
         return Is("nv_platform");
     }
+    public static bool IsNVPlatformExists()
+    {
+        return Exists("nv_platform");
+    }
 
     public static bool IsForceSetGPUMode()
     {
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -664,9 +664,9 @@ public static class AppConfig
         return Is("shutdown_reset") || ContainsModel("FX507Z");
     }
 
-    public static bool IsNVServiceRestart()
+    public static bool IsNVPlatform()
     {
-        return Is("nv_restart");
+        return Is("nv_platform");
     }
 
     public static bool IsForceSetGPUMode()
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -1632,7 +1632,7 @@ namespace GHelper
             checkNVPlatform.Padding = new Padding(3);
             checkNVPlatform.Size = new Size(917, 42);
             checkNVPlatform.TabIndex = 15;
-            checkNVPlatform.Text = "Stop NV Platform in Eco";
+            checkNVPlatform.Text = "Stop NVidia Services in Eco mode";
             checkNVPlatform.UseVisualStyleBackColor = true;
             // 
             // Extra
--- a/app/Extra.Designer.cs
+++ b/app/Extra.Designer.cs
@@ -120,6 +120,7 @@ namespace GHelper
             checkVariBright = new CheckBox();
             checkGpuApps = new CheckBox();
             checkGPUFix = new CheckBox();
+            checkOptimalBrightness = new CheckBox();
             checkStatusLed = new CheckBox();
             checkPerKeyRGB = new CheckBox();
             panelPower = new Panel();
@@ -143,7 +144,7 @@ namespace GHelper
             buttonACPISend = new RButton();
             pictureDebug = new PictureBox();
             labelACPITitle = new Label();
-            checkOptimalBrightness = new CheckBox();
+            checkNVPlatform = new CheckBox();
             panelServices.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)pictureService).BeginInit();
             panelBindingsHeader.SuspendLayout();
@@ -182,18 +183,18 @@ namespace GHelper
             panelServices.Controls.Add(labelServices);
             panelServices.Controls.Add(buttonServices);
             panelServices.Dock = DockStyle.Top;
-            panelServices.Location = new Point(13, 1616);
+            panelServices.Location = new Point(15, 1806);
             panelServices.Name = "panelServices";
-            panelServices.Size = new Size(830, 66);
+            panelServices.Size = new Size(949, 75);
             panelServices.TabIndex = 5;
             // 
             // pictureService
             // 
             pictureService.BackgroundImage = Resources.icons8_automation_32;
             pictureService.BackgroundImageLayout = ImageLayout.Zoom;
-            pictureService.Location = new Point(18, 17);
+            pictureService.Location = new Point(21, 19);
             pictureService.Name = "pictureService";
-            pictureService.Size = new Size(28, 28);
+            pictureService.Size = new Size(32, 32);
             pictureService.TabIndex = 21;
             pictureService.TabStop = false;
             // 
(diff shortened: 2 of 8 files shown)


zzoko:
not sure if this issue is expected but i got around to trying this feature and this is what i noticed:

if g14 is turned off while being in Eco mode without "Enable GPU on shutdown" enabled after you start up the pc again when you switch from Eco mode to Standard the screen will completely freeze and need to be hard rebooted with the power button

1. im pretty sure its just the screen not actual pc (because keyboard function keys still work and i can see my windows hello camera light if i press ctrl + l)
2. in nvidia control panel i use dedicated gpu only, so Eco mode is igpu only and Standard mode is dgpu only, i have not tested it with optimus/advanced optimus/ultimate mode so the behavior may be different for that
3. using g14 2025 model rtx 5070 

if "Enable GPU on shutdown" and "Stop NVidia Services in Eco mode) are both enabled then the feature works as expected with no issues, with the only noticeable difference being a very brief black screen when switching to Eco mode, and ofc the nvidia services get stopped)

not really an issue for me because i use "Enable GPU on shutdown" anyway to also fix the brightness bug, but others that don't have that enabled may get confused by the freeze

also theres a wrong capital letter in "NVidia" :P

seerge:
@zzoko hello,

It's not expected, but it's also not due to G-Helper's actions (you are free to verify that without app running even) :) 

It looks more like an actual Nvidia Advanced Optimus acting and usually it can be solved by doing a [clean Nvidia driver reinstall using DDU](https://github.com/seerge/g-helper/wiki/Troubleshooting#reinstalling-gpu-drivers) (in safe mode ideally). Or by setting Nvidia Optimus state to Auto or actual Optimus.

> if "Enable GPU on shutdown" and "Stop NVidia Services in Eco mode) are both enabled then the feature works as expected with no issues, with the only noticeable difference being a very brief black screen when switching to Eco mode, and ofc the nvidia services get stopped)

Does checking ONLY "Stop NVidia Services in Eco mode" help with that ? or is booting in Standard ("Enable GPU on shutdown") is needed for you as well ?

As for Nvidia naming - thanks for the tip, i'm kind of lost since they seem to call themselves NVIDIA (all caps) now. I probably need to rephrase the whole checkbox text at all, since it's not only stopping but also starting NV services as well (when enabling GPU of course)

zzoko:
to clarify
1. if both "Stop Nvidia Services in Eco mode"  and "Enable GPU on shutdown" are **off** everything works as expected, the only minor problem is that screen brightness can't be changed if laptop was shutdown while in Eco mode 

2. if only "Enable GPU on shutdown" is **on**, everything works as expected, also no brightness issue

4. if only  "Stop Nvidia Services in Eco mode" is **on** this new bug appears, where if you shut down the laptop on Eco then turned it back on if you then switched from Eco to Standard mode the screen will freeze 

4. if both "Stop Nvidia Services in Eco mode"  and "Enable GPU on shutdown" are **on** everything works as expected

my theory is perhaps the screen runs on iGPU then when i set to standard mode the screen tries to run on dGPU but because it switches before the services have enough time to start the screen just freezes? perhaps adding a short delay between starting the nvidia services and triggering the gpu switch?

to sum it up the feature itself works without issue, i can go between Eco and Standard as much as i want and all services start/stop fine, the funky issue seems to come if laptop was shut down in Eco without Enable GPU on shutdown

seerge:
@zzoko thanks for a clarification!

For (3) - the thing is that `Stop NV services` doesn't disable them from startup, it just stops them the moment you select Eco. So when you will reboot (in Eco mode) services will still start next time with windows.

So from technical standpoint it's should be exactly same as if you would just do nothing (1) :| 

Is that situation repeatable ? Or happened only once ? Does it matter in what specific state Nvidia Display Mode (Optimus / Nvidia GPU only) was before setting Eco ?

zzoko:
It was repeatable every time if i use option 3. from last comment.

I went ahead and tried using old optimus mode instead of dGPU only mode in nvidia control panel and the issue did not happen. (Kinda expected that)

After testing that i switched back to dGPU mode only in control panel (since its marginally better performance and gsync support).

Im fairly certain you should be able to replicate the screen lockup since the pc and windows install are basicall new:
1. Go to Display mode in nvidia panel and set it to NVIDIA GPU only.
image
2. Restart your pc (important since if you are were using optimus previously the issue doesn't happen).

3. Enable "Stop Nvidia Services in Eco mode" and leave "Enable GPU on shutdown" off.
4. Set ghelper to Eco.
5. Turn off pc.
6. Start pc and switch from Eco to Standard mode, when it tries to switch screen from iGPU to dGPU you will see the nvidia notification for display now switched to dGPU then the screen will lock up.


If you are able to replicate the issue a band aid fix would be when this new feature is enabled if someone is in dGPU display mode to force enable "Enable GPU on shutdown".

seerge:
@zzoko thanks for a detailed explanation! 

I think I have managed to reproduce it once (even tho it doesn't happen to me consistently).

Reason is likely not the fact that NV services are stopped in `Eco`, but that they are **(re)**started when `Standard` is toggled.  And in this particular case restarted too fast. My assumption this was happening exactly in the middle of Nvidia services setting optimus state to Nvidia GPU only :)

Try this build with a delay and let me know if it behaves differently!

[GHelper.zip](https://github.com/user-attachments/files/22857626/GHelper.zip)

zzoko:
Hm issue still happens the same way, i took a vid replicating the issue here https://drive.google.com/file/d/18TvvM6_uEnnP7En0yaaGYVqyYaYK6I37/view?usp=sharing

-------

Side note the GPU _core clock offset_, _memory clock offset_ and _core clock limit_ sliders do no save per profile unlike all other sliders for cpu & gpu, and instead are applied globally to all profiles, is that the way its intended to be or a bug? https://drive.google.com/file/d/1roWhj3xuiRBoR5z72zdEcZm4CWSGPP9y/view?usp=sharing

Update: for the GPU clocks issue i read your comment [here](https://github.com/seerge/g-helper/issues/3669#issuecomment-2623818148) and while i understand why you did it i think it's not a good solution and should be made obvious in some way that manually setting it to 0mhz and clicking factory defaults to set it to 0mhz will make the slider behave differently when switching profiles without needing to dig up github comments

seerge:
@zzoko I have extended delay to 5 seconds, try this build instead

[GHelper.zip](https://github.com/user-attachments/files/22863433/GHelper.zip)

If it still happens, try the following
1. Check `Stop NV services`
2. Set `Eco` and shutdown
3. **Uncheck** `Stop NV services`
4. Set `Standard`

Does it still happen ? Cause in this scenario G-Helper won't be (re)starting services at all when `Standard` is set.

zzoko:
Just tested the version you sent and can confirm it works, no screen freeze when switching to Standard mode, the Stop NV Services also fixes the brightness issue so that's great. 
I hope you do also do something about the GPU slider behavior though even a tooltip when hovering over it would suffice.

Otherwise great job your app is stellar :D

seerge:
@zzoko glad to hear!

As for Clock offsets slider it's absolutely a separate topic.

Thing is that most of the params that you can set are just asus-proprietary and firmware based, and if they are not explicitly applied - firmware uses some default values that can't be even read. This is CPU power limits, GPU power/boost, GPU temp limits, etc (pretty much everything you have in AC's Manual mode).

Default state for them is not to be applied (which is same as in AC) and it fits well in my G-Helper's concept to not to touch anything, unless user explicitly have chosen to change them.

----

Only a few are non-asus specific. Namely CPU boost (windows power plan setting) + GPU clock offsets (Nvidia driver). They can be read and displayed. So here my approach is also to "do nothing" by default and just show currently active value.

I was considering to change UI a bit, like making sliders greyed-out if they are not active (same as i did with fan curves recently).

Scarlaid:
I know this is a pull request

but bless you for thinking about this, because this DRASTICALLY improved battery life
though I am curious, is this the final nail for battery optimization or there are still tricks up the sleeves?

Issue #4575, opened 2025-09-18
Adding Ambient light-controlled keyboard backlight using WinRT library


dcx07:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information will be provided, my issue will be closed without an answer.

### Is your feature request related to a problem? Please describe

No, it isn't, but it would be nice to add this feature since some Asus laptops （mostly Zenbooks）have this feature with the AsusOptimization service enabled.

### Describe the solution you'd like

Adding feature for reading and controlling the keyboard backlight based on ambient light sensor readings.

### Describe alternatives you've considered

Keep AsusOptimization running with g-helper, with the background process AsusOSD.exe killed to avoid confilcting with g-helper's OSD, but AsusOptimization conflicts with g-helper on the fan settings and keyboard backlight levels.

### Device and Model

Asus Zenbook UX5406

### Additional information.

I wrote a very simple [demo](https://github.com/dcx07/LightSensor) here to show how you can read ambient light sensor values.

xaviergonz:
@seerge Guess I didn't need to add a new issue in the end :) +1 to this feature

Some other example code
```cs
using Windows.Devices.Sensors;

static double? ReadAmbientLux()
{
    LightSensor? sensor = LightSensor.GetDefault();
    LightSensorReading? reading = sensor?.GetCurrentReading();

    return reading?.IlluminanceInLux;
}

static LightSensor? StartAmbientLuxChanged(Action<double> onLuxChanged)
{
    LightSensor? sensor = LightSensor.GetDefault();
    if (sensor is null)
        return null;

    sensor.ReportInterval = Math.Max(sensor.MinimumReportInterval, 250u);

    sensor.ReadingChanged += (_, args) =>
    {
        double lux = args.Reading.IlluminanceInLux;
        onLuxChanged(lux);
    };

    return sensor; // keep this alive while you want events
}
```

xaviergonz:
Copying post here so it can be tracked in the proper issue.

I did a small proof of concept for this, and it looks like it is possible to read the ambient light sensor without WinRT :D

[...]

PR #4577, opened 2025-09-18
feat: Added Ambient light sensor-controlled keyboard backlight functionalities

dcx07:
Proposed solution for #4575, but optimization needed


dcx07:
Finished feature adding, but I do not know why I'm getting 200MB of compiled size.

PR #4674, opened 2025-10-14
Add priority setting

Iemand005:
Added automatically setting the process priority to normal on startup.

This can be expanded on by allowing the user to configure GHelper to automatically launch with a priority of choice.

This is necessary as the scheduled task runs it at below normal priority to prevent it from being the first to stop responding in case of CPU bottleneck.

### DIFF
--- a/app/Helpers/ProcessHelper.cs
+++ b/app/Helpers/ProcessHelper.cs
@@ -145,6 +145,19 @@ namespace GHelper.Helpers
             return result;
         }
 
+        public static void SetPriority(ProcessPriorityClass priorityClass = ProcessPriorityClass.Normal)
+        {
+            try
+            {
+                using (Process p = Process.GetCurrentProcess())
+                    p.PriorityClass = priorityClass;
+            }
+            catch (Exception ex)
+            {
+                Logger.WriteLine(ex.ToString());
+            }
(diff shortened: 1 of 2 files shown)


seerge:
@Iemand005 hello, that's reasonable :) Merged

Iemand005:
Awesome, thank you very much!

seerge:
@Iemand005 do I understand correctly that it fixes your problem with app being responsive in extreme situations (like 5W limit + extra load from system / other apps)

Iemand005:
That's correct, that's why I added it.

As far as I can tell you can request up to High priority level without admin rights. Realtime requires admin.

I set it to normal as you said you don't want GHelper to be prioritized over other processes, which I agree with, it should at least not be the first thing that stops responding anymore which is important as you need GHelper to respond to get out of a low power state.

Having GHelper request Normal priority on startup should always make sure it's run as a normal process even if the task scheduler wants to run it on low priority.  It's possible you can configure the scheduled task to run it at normal priority too but that's a bit more of a mess. This solves my performance problems.

If I ever have issues with it becoming unresponsive I might consider adding an extra setting to let the user configure GHelper to automatically switch to a higher priority level by default. This'll do for now though.

PR #4940, opened 2026-01-10
Add support for ROG Harpe II Ace (Motion Sync & Zone Mode)

Benson5650:
## Description
This PR adds full support for the **ROG Harpe II Ace**, implementing its specific features including **Motion Sync** and **Zone Mode**.

Verified on my ROG Harpe II Ace.
image

### DIFF
--- a/app/Properties/Strings.Designer.cs
+++ b/app/Properties/Strings.Designer.cs
@@ -8,11 +8,10 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace GHelper.Properties
-{
+namespace GHelper.Properties {
     using System;
-
-
+    
+    
     /// <summary>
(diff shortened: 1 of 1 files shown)


seerge:
@Benson5650 hello, looks good! Kudos for figuring out how it works and reverse-engineering 2 new settings.

I don't have ROG Harpe II Ace by myself to test, but i assume it works. So merged

Thanks!

Benson5650:
I realized that my recent changes broke backward compatibility for mouse profile imports (`.gmp1` files). The profile format now includes 2 additional bytes, causing older exported profiles to fail the length validation during import.

Should we maintain backward compatibility, or I can just remove these byte because saving motionsync and zonemode is not that necessary.

seerge:
@Benson5650 hello, let's just revert that then.

Issue #4953, opened 2026-01-15
"Power Mode StatusChange: Online" spam with non-passthrough usb-c charging that makes g-helper go completely unresponsive in some cases


EminGT:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

When using a laptop that doesn't have passthrough usb-c charging but instead has regular usb-c charging like the 2025 Zephyrus G14 with a 5070ti or above, G-Helper constantly gets "Power Mode StatusChange: Online" events if there is a set low charge limit. This makes the program go completely unresponsive if the charger_delay value is high enough. In my experience, setting charger_delay to anything higher than 2000 is enough to get this issue to make g-helper unusable.

### How to reproduce the bug?

1. Get a laptop that has non-passthrough charging like the Zephyrus G14 with a 5070Ti
2. Set a charger_delay value of 5000
3. Plug in a usb-c charger
4. Set charge limit to 60%
5. Wait until charge hits exactly 59.5% where the issue triggers
6. Observe that G-helper basically freezes and gets completely unusable

### Logs

[log.txt](https://github.com/user-attachments/files/24653422/log.txt)

### Device and Model

Zephyrus G14 GA403WW

### Additional information.

The problem might seem like an edge case, but using a high charger delay is a must for some people with weak chargers or weak electricity, and using a 60% battery limit is the most healthy way to use non-passthrough usb-c charging. This problem also makes usb-c charging consume more power than usual due to G-Helper constantly doing stuff in the background, and this makes the use of laptop charging powerbanks really inefficient as well. I have measured around 2-3 watts of wasted power compared to shutting down g-helper while usb-c charging.

### Armoury Crate

Uninstalled

### Asus Services

None

### Version

Happens on both v0.237 and v0.238

### OS

Win 11 25H2

seerge:
@EminGT hello,

Power mode change comes from windows (and window gets it from hardware), app just logs it (and reacts accordingly if it changes).

System should not be constantly switching between battery / plugged in by itself at all. If it happens - it's some underlying hardware thing.

[...]

PR #4954, opened 2026-01-16
non-passthrough fix (issue #4953)

EminGT:
I just reordered the SystemEvents_PowerModeChanged function with two key changes:

1) Added an extra earlier "isPlugged ? then return" check to catch Windows triggering power mode change when it's still plugged
2) Move the Power Mode logger to a line after that, so that these random triggers don't spam the log all the time and cause unnecessary disk writes, consuming power
3) -Bonus- The "disable_power_event ? return" check is now done earlier as well because why not

### DIFF
--- a/app/Program.cs
+++ b/app/Program.cs
@@ -305,7 +305,6 @@ namespace GHelper
 
         private static void SystemEvents_PowerModeChanged(object sender, PowerModeChangedEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


EminGT:
Fixed in commit [8ee47ab](https://github.com/seerge/g-helper/commit/8ee47ab8eed0e541a732ca20980562e76e2522bf)

Issue #4957, opened 2026-01-16
DPI Profile Count Reduced When Switching Profiles via G-Helper


Benson5650:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

Usually, we have 4 DPI profiles. But Asus Gear Link actually allow fewer Dpi profiles (my mouse doesn't support armoury crate, not sure if AR or other mice has similar function.)

<img width="566" height="377" alt="Image" src="https://github.com/user-attachments/assets/4e3b7889-f60f-4eee-8978-e589c19d350e" />

Assume that I have 4 profiles at the beginning.

<img width="561" height="390" alt="Image" src="https://github.com/user-attachments/assets/3394de95-4250-414e-ac2c-12a0425503a3" />

When I set a dpi profile in G-helper (take DPI 2 as an example)

<img width="457" height="234" alt="Image" src="https://github.com/user-attachments/assets/50526fe9-4f93-4ed0-9aa1-1366c44be1fc" />

There will be only 2 profiles left.

<img width="579" height="392" alt="Image" src="https://github.com/user-attachments/assets/ebf74a90-c276-409d-a016-54b3f8759b97" />

as well as setting to first profile .

<img width="463" height="236" alt="Image" src="https://github.com/user-attachments/assets/33af401d-830d-408a-a8fe-5b5c295e2ec6" />

<img width="584" height="380" alt="Image" src="https://github.com/user-attachments/assets/3d9b587c-6d47-4d2f-863f-697cc8595851" />

I'm not sure that if any other mouse has this issue, maybe harpe ii ace has different protocol?

It seems to be related to the usage in app\Peripherals\Mouse\AsusMouse.cs L1257~L1261:
```
            //The first DPI profile is 1
            WriteForResponse(GetChangeDPIProfilePacket(profile));
            //For whatever reason that is required or the mouse will not store the change and reverts once you power it off.
            WriteForResponse(GetChangeDPIProfilePacket2(profile));
            FlushSettings();
```

[...]

PR #4965, opened 2026-01-18
Dpi profile fix for supported mouse

Benson5650:
#4957 stiil working on this, not sure if this will brake other mice.


Benson5650:
- For those mice do not support dynamic DPI count, the only difference is SetDPIProfile nolonger send             `WriteForResponse(GetChangeDPIProfilePacket(profile));`
This should be fine because `WriteForResponse(GetChangeDPIProfilePacket2(profile));` is the correct one base on my observation. (**need more test on different device**)

- Also implemented many UI fix, since most of the objects rendered dynamically we not be resized.

- add new mouse to supported list.

seerge:
@Benson5650 hello, great!

It turns out my Harpe Ace Mini also supports custom number of profiles (I have set flag to true for it and it works). Good find!

Merging.

Issue #4806, opened 2025-12-01
GHelper sees logo and cover lighting on devices that don't have them.


MinecraftNerrd1337:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

I have a ROG G16 G614PM, and when I go to the section where you can alter lighting behavior according to system state (ie. lightbar during startup), it sees four integrated lighting peripherals, but my laptop only has two, and I believe this is the case for all G614PMs (I only know of two SKUs).
Here's a screenshot (my system is in French):
<img width="741" height="321" alt="Image" src="https://github.com/user-attachments/assets/6b4844e6-4556-41f1-b9a2-a6280b7c5bb4" />
I should only be seeing keyboard (clavier) and lightbar (barre lumineuse). I found this to also be the case with another laptop, I believe a G713RM, but I no longer own that device.
For models that have SKUs with and without logo or cover lighting, maybe adding an option to configure which of them actually exists in the json (power user settings) would be a good idea.

### How to reproduce the bug?

1. Install GHelper on your ROG Strix G614PM (likely also applies to a lot of similar Strix laptops)
2. On main Window, under "keyboard", click on "more options"
3. In the appearing Window, right under "keyboard backlight", notice that there are lighting behavior settings for four integrated lighting peripherals, despite the device only having two that exist.

### Logs

[log.txt](https://github.com/user-attachments/files/23860658/log.txt)

### Device and Model

Asus ROG Strix G614PM

### Armoury Crate

Uninstalled

### Asus Services

5 (MyAsus related, doesn't affect my issue)

### Version

0.233.0

### OS

Windows 11 25H2 (doesn't affect my issue)

seerge:
@MinecraftNerrd1337 hello,

Do I understand that Ambient / Heatmap modes also "flicker" in your case too same as for this guy https://github.com/seerge/g-helper/issues/4719 ?

If so , your device has only 4 zone rgb (and as it usually happens doesn't have extra lighting zones). I just need to add it to the list.

seerge:
[...]

Issue #4719, opened 2025-10-25
Ambient Backlight mode flickering


fadi9711:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

When applying Ambient Backlight mode, it flickers.
I had my old laptop, ROG Strix G15 G513RM, and it was working perfectly. Now I bought a new laptop ROG Strix G16 G614PR, and the back lights of the keyboard will flicker when I use Ambient mod (the lights of the lid bar will go off)

### How to reproduce the bug?

1. Open G-helper
2. Laptop Keyboard 
3. Choose Ambient 

### Logs

[log.txt](https://github.com/user-attachments/files/23144308/log.txt)

### Device and Model

Asus Rog Strix G16 G614PR

### Additional information.

I am trying to upload a video, but it says failed every time.

### Armoury Crate

Uninstalled

### Asus Services

None

### Version

0.234.0

### OS

Windows 11 25H2

D0minatorX:
[6046579998691694832.zip](https://github.com/user-attachments/files/23144786/6046579998691694832.zip)

This is a video record for the issue I made it as zip so I can upload it here

seerge:
@fadi9711 hello,

Do you see Per-key RGB toggle under Extra settings? If so can you tick it and see if it helps?

D0minatorX:
@seerge Hi

No I don't have this option.

Thank you for your reply

seerge:
@fadi9711 then I need to add your device to the list where this setting is possible 

I'm away from the laptop atm, so you will have to wait like couple of weeks :)

D0minatorX:
@seerge 

Ok, sir no problem. I will wait.

There is another thing: when I apply a custom fan curve, it loads the CPU. I do not know how to explain it in a better way.
For example, the temps will be 60 degrees, and when I apply the curve, it jumps to 80.

Thanks for your help

seerge:
@fadi9711 hello, can you try this build and let me know if it behaves differently here ?

[GHelper.zip](https://github.com/user-attachments/files/23797920/GHelper.zip)

Thanks

[...]

PR #4981, opened 2026-01-25
AppConfig: Add G614PP to IsStrixLimitedRGB()

LeddaZ:
Same situation as https://github.com/seerge/g-helper/issues/4806 and https://github.com/seerge/g-helper/issues/4719


seerge:
@LeddaZ that's valid :) Merged

PR #5043, opened 2026-02-14
AppConfig: Add G615LR to IsStrixLimitedRGB()

LeddaZ:
This model has per-key RGB but no logo/slash

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -520,7 +520,7 @@ public static class AppConfig
 
     public static bool IsStrixLimitedRGB()
     {
-        return ContainsModel("G614PM") || ContainsModel("G614PP") || ContainsModel("G614PR") || ContainsModel("G512LI") || ContainsModel("G513R") || ContainsModel("G713QM") || ContainsModel("G713PV") || ContainsModel("G513IE") || ContainsModel("G713RC") || ContainsModel("G713IC") || ContainsModel("G713PU") || ContainsModel("G513QM") || ContainsModel("G513QC") || ContainsModel("G531G") || ContainsModel("G615JMR") || ContainsModel("G615LM") || ContainsModel("G815LR");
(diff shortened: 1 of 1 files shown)


seerge:
@LeddaZ that's valid :) Merged.

HHDSORA:
@seerge My laptop model is also the G615LR, but it has Logo RGB lighting.

LeddaZ:
Screenshot_2026-03-05-07-37-41-50_40deb401b9ffe8e1df2f1cc5ba480b12

Hm, it seems that Asus made the same exact laptop 2 times, with the only difference being the logo. I'm not sure if there's a way to distinguish between them in software.

seerge:
@HHDSORA you can actually enable per-key RGB back under `Extra` settings (just restart app after that). Does it solve the problem for you ?

HHDSORA:
@seerge No, I tried doing that, but there are still only options for the keyboard RGB and the bottom light strip.

seerge:
@HHDSORA @LeddaZ  ok, then I guess I need to remove this model from the list

Do I understand correctly that for both of you device has per-key RGB (just in one case it also has logo / lightbar and in other it does not ?)

If so - there is no harm to keep logo / bar checkboxes under `Extra` available, they will just have no effect if there are no physical logo of course :)

This build should do it

[GHelper.zip](https://github.com/user-attachments/files/25767500/GHelper.zip)

LeddaZ:
> @HHDSORA @LeddaZ ok, then I guess I need to remove this model from the list
> 
> Do I understand correctly that for both of you device has per-key RGB (just in one case it also has logo / lightbar and in other it does not ?)
> 
> If so - there is no harm to keep logo / bar checkboxes under `Extra` available, they will just have no effect if there are no physical logo of course :)
> 
> This build should do it
> 
> [GHelper.zip](https://github.com/user-attachments/files/25767500/GHelper.zip)

Yeah, one of them has the logo light and the other doesn't, and they both have lightbar and per key RGB

HHDSORA:
@seerge It's working properly now, thank you

Issue #4977, opened 2026-01-23
Color/display problem


p35p361412-glitch:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

display/color is a bit off, it's look like the "mist" appers on usual colors, only with HDR on or automatic manage color (AMC) on it stabilized.

![Image](https://github.com/user-attachments/assets/13e275ce-f9d2-4f2e-8fe9-de97294f97dc)

![Image](https://github.com/user-attachments/assets/2007c4fb-18b1-4dae-8928-a6514b7816b1)

![Image](https://github.com/user-attachments/assets/aee7a154-1ccd-4457-ac9d-4505eb3b2ff6)

Original color
<img width="2190" height="1706" alt="Image" src="https://github.com/user-attachments/assets/788c28c3-5600-419b-b223-27a6c333a404" />

The issue
<img width="2879" height="1707" alt="Image" src="https://github.com/user-attachments/assets/5e485cc8-f2a5-4b57-8637-48528de68d6a" />

### How to reproduce the bug?

Just started the G14 and its appears 

### Logs

its started about 23.01.26 14:00-15:00, before that it was still normal

[log ghelper.txt](https://github.com/user-attachments/files/24822668/log.ghelper.txt)

### Device and Model

Asus zephyrus G14 GA403UV

### Additional information.

I tried switching back and forth HDR, ACM, flicker-free, gamut, cold and warm, gpu mode in eco and standart, 60/120 hz, restarting - its not helping 
I didn't try to calibrate display color because of warning 

<img width="2877" height="1703" alt="Image" src="https://github.com/user-attachments/assets/1381769a-081b-41ce-8d7e-4ec35c359697" />

### Armoury Crate

Uninstalled

### Asus Services

0

### Version

0.237.0

### OS

Asus zephyrus g14 ga403uv

seerge:
@p35p361412-glitch hello, I don't think that G-Helper or Asus color profiles (handled by their own utility) are actually related to this.

But for a sanity check try to do following

[...]

PR #5048, opened 2026-02-16
AMD OLED Power Optimization causing washed colors alert fix

spherohero:
Issue #4977 is fixed by adding a preventative pop up that appears when on battery, SDR, and on app open to notify users to disable AMD OLED power optimization. Popup gives users prompt to close regularly or close and not reopen. 

*Add to wiki in troubleshooting section rather than just where it was placed in README.md

### DIFF
--- a/app/Settings.cs
+++ b/app/Settings.cs
@@ -484,6 +484,34 @@ namespace GHelper
             VisualiseDisabled();
         }
 
+        private void OLEDHDRWashWarning(bool hdr)
+        {
+            if (AppConfig.Is("oled_hdr_wash_warning_shown")) return;
+            if (!AppConfig.IsOLED()) // varibright affects OLED panels
+            {
+                return;
+            }
+            if (SystemInformation.PowerStatus.PowerLineStatus != PowerLineStatus.Offline) // on battery power
+            {
+                return;
+            }
+            if (!(AppConfig.IsG14AMD() || AppConfig.IsAMDLight() || AppConfig.IsAMDiGPU())) // AMD varibright only on AMD hardware
+            {
+                return;
+            }
+            if (hdr) // varibright only affects when on SDR
+            {
+                return;
+            }
+
+            string message = "On battery power and using SDR, OLED screen may be washed out due to AMD OLED Power Optimization.\n\nPlease disable OLED Power Optimization in AMD Software: Adrenalin Edition -> Settings -> Display -> OLED Power Optimization.\n\nClick Cancel to not show this message again.";
+
+            if (MessageBox.Show(message, "OLED Color Issue", MessageBoxButtons.OKCancel, MessageBoxIcon.Warning) == DialogResult.Cancel)
+            {
+                AppConfig.Set("oled_hdr_wash_warning_shown", 1);
+            }
+        }
+
         public void VisualiseBrightness()
         {
             Invoke(delegate
@@ -1293,6 +1321,8 @@ namespace GHelper
 
         public void VisualiseScreen(bool screenEnabled, bool screenAuto, int frequency, int maxFrequency, int overdrive, bool overdriveSetting, int miniled1, int miniled2, bool hdr, int fhd, int hdrControl)
         {
+
+            OLEDHDRWashWarning(hdr);
 
             ButtonEnabled(button60Hz, screenEnabled);
             ButtonEnabled(button120Hz, screenEnabled);
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,6 +24,7 @@ Small and lightweight Armoury Crate alternative for Asus laptops offering almost
 - [FAQ](https://github.com/seerge/g-helper/wiki/FAQ)
 - [Setup and Requirements](https://github.com/seerge/g-helper/wiki/Requirements)
 - [Troubleshooting](https://github.com/seerge/g-helper/wiki/Troubleshooting)
+  - *Note: If you experience washed out colors on OLED screens on battery while using SDR, disable AMD OLED Power Optimization in AMD Software.*
 - [Power User Settings](https://github.com/seerge/g-helper/wiki/Power-user-settings)
 
 


seerge:
@spherohero hello,

This is a good idea to notify, but I'm not sure about current way of checking you propose :) Throwing a popup for ALL AMD OLED devices is likely too much.

I have tested, AMD Adrenaline saves state of OLED Optimization under (at least on my laptop)
```
HKLM\System\CurrentControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}\0001\DAL2_DATA__2_0\DisplayPath_0\EDID_834C_419C\Adjustment\DAL_SCE_Settings
```
But I'm not sure if it's universal.

seerge:
@spherohero and the most important, i have latest 26.2.1 driver, and even on battery OLED Optimization doesn't wash out colors anymore :)

Can you check if this is the case for you ?

spherohero:
I haven't checked the saved state but I will so in a bit, the issue with the 26.2.1 driver is this update is still not on my laptop with it saying that the latest version is still 26.1.1 as of last checked today. (With this the colors do still wash out on battery OLED optimization)

seerge:
@spherohero try to get latest from here https://videocardz.com/driver/amd-radeon-software-adrenalin-26-2-1 :)

seerge:
@spherohero do i understand correctly that 26.2 fixes it ?

spherohero:
Accidentally renamed the branch and it auto deleted it oops, the update is still downloading my internet speed is not the fastest.

spherohero:
I tested it again on 26.2.1 and at first it worked correctly with the colors not washing out immediately, but if you change the brightness via g-helpers slider, it instantly washes out once again.

ItzJoseD:
@seerge 

I experienced this exact issue and here is how it occurred:

I enabled OLED power optimization and disabled vari-bright, then I disabled "automatically manage color for apps", and then I when into G-Helper and switched the display from Native to sRGB. 

Once i did that it washed out the colors on the display and made it look extremely white like brightness had been cranked. I will say, once it is in this bugged state it seems to stick to this bugged state and I could only get it to revert by using DDU and re-installing AMD graphics drivers.

Its better than the older AMD drivers which washed out immediately, compared to these that only wash out AFTER you switch the display's color gamut. 

It doesnt seem to be a problem with G-helper but I can confirm on my Asus G14 2024.

I also want to confirm the setting for AMD's OLED optimization is indeed saved under "Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}\0001\DAL2_DATA__2_0\DisplayPath_0\EDID_834C_419C\Adjustment".

That is where DAL_SCE_Settings is saved and is where the value "00 00 00 00 02 00 00 00" lies, 02 occurs when AMD's OLED power optimization feature is enabled and seems consistent across multiple platforms.

seerge:
@LunarVGaming1 ok, that's interesting, considering it's main display likely DisplayPath will be same.

@spherohero can you check if reg path is same for your device ? I.e.
```
HKLM\System\CurrentControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}\0001\DAL2_DATA__2_0\DisplayPath_0\EDID_834C_419C\Adjustment\DAL_SCE_Settings
```

spherohero:
Yup can confirm that it changes to 02 
on the same path when OLED power optimization is on. However to replicate the issue I don't need to change G-Helpers color gamut to anything else. On native the issue occurs instantly and I can revert back to regular colors just by disabling the power optimization setting

seerge:
@spherohero @LunarVGaming1 

Ok, then I would actually check for this flag then. I can imagine EDID_<xx> part could be different for other models, so i have made if flexible.

I have added the check before actually running any AsusSplendid.exe command, and if it's true (+ it's AMD OLED on battery) - it will entirely skip it (and mention in the log)

[GHelper.zip](https://github.com/user-attachments/files/25496368/GHelper.zip)

seerge:
@spherohero @LunarVGaming1 one more update, I have added a button in the bottom right saying AMD OledSaver that opens Adrenaline if clicked (i plan to gather all "strange" notifications in same area) 

It appears when AsusSplendid call is cancelled due to flag beging ON (which would happen on app start or attempt to change any Visual setting)

[GHelper.zip](https://github.com/user-attachments/files/25496908/GHelper.zip)

spherohero:
Tested this new change, the button only appears after adjusting the flicker-free dimming or visual mode, not automatically on app start. The button also does not disappear if visual mode is changed, it requires a full restart of G-helper app. Also I don't know if the change occurred for you guys, but OLED power optimization also washes out colors when plugged in now as well.

seerge:
@spherohero likely you had Default visual mode ? It's not reapplied on start - so no splendid call, no check

I have added a flag check on form creation (i.e. app start) no matter what 
And now it will check for a flag status after button click and hide it on next form focus (same logic applies to `Energy Saver` status checker)
Also flag will ignore battery status (i.e. will fire true if plugged in too)

[GHelper.zip](https://github.com/user-attachments/files/25499852/GHelper.zip)

spherohero:
Works perfectly now!

I had tried before by both changing flicker-free dimming and visual modes, but now it prompts and disappears when appropriate.

Thanks

seerge:
@spherohero @LunarVGaming1 ok,  great

I did some optimizations to make this check run even faster 

Can you check if this build still detects it correctly - if so I assume it's solved

[GHelper.zip](https://github.com/user-attachments/files/25542685/GHelper.zip)

ItzJoseD:
@seerge 
It worked fine but only if G-helper is initially launched after OLED power optimization is enabled, if it is enabled after launching it is not detected and the button does not show up.

other than that it does work perfectly, and it launches AMD adrenaline when pressed to allow the disabling of the feature, thanks for adding that!

seerge:
@ItzJoseD that's very exotic situation :)

Checking for this setting, is quite resource heavy, so I have added it on start + every time you actually change Visual mode / flicker-free dimming / etc (i.e. run AsusSplendid.exe under the hood). 

That should be enough.

ItzJoseD:
@seerge 
Ah okay perfect, ill make sure that functionality you mentioned works, I hadnt tested the flicker free dimming triggering detection. Thanks!

PR #5159, opened 2026-03-27
Add ROG Strix Go 2.4 headset support

### LINKED ISSUE #5158: Add ROG Strix Go 2.4 headset support
## Summary
- Add initial support for the ROG Strix Go 2.4 wireless headset in the peripherals panel
- Detect the headset through the parent MI_03 HID interface and read battery through feature reports
- Use a headset icon and show battery status in G-Helper

## Device Details
- VID: 
- PID: 
- Feature report interface: parent 
- Battery request packet: 

## Remaining Gaps
- Detect headset power-off state reliably and update the UI to 
- Add a dedicated headset settings window on click
- Future settings could include volume-related controls, sound modes, and auto power-off timeout similar to Armoury Crate

## Validation
- Headset appears in the peripherals panel
- Battery reads successfully on the reporter's device
- Dongle disconnect removes the tile and reconnect restores it

vancinis:
Closes #5158

## Summary
- add initial support for the ROG Strix Go 2.4 wireless headset in the peripherals panel
- detect the headset through the parent MI_03 HID interface and read battery through feature reports
- use a headset icon and show battery status in G-Helper

## Changes
| File | Change |
|------|--------|
| `app/Peripherals/Headset/AsusHeadset.cs` | Added a headset peripheral base class using HID feature reports |
| `app/Peripherals/Headset/Models/StrixGo24.cs` | Added the ROG Strix Go 2.4 model and battery request parsing |
| `app/Peripherals/PeripheralsProvider.cs` | Added headset discovery, connection, disconnection, and battery refresh integration |
| `app/Peripherals/IPeripheral.cs` | Added the `Headset` peripheral type |
| `app/Program.cs` | Added headset detection during startup |
| `app/Settings.cs` | Added headset tile rendering and headset icon usage |
| `app/Properties/Resources.resx` | Registered the new headset icon resource |
| `app/Properties/Resources.Designer.cs` | Exposed the headset icon resource property |
| `app/AnimeMatrix/Communication/Platform/WindowsUsbProvider.cs` | Fixed `Get()` to return the feature report response buffer |

## Test Plan
- [x] `dotnet build app/GHelper.sln`
- [x] Manual testing with a real ROG Strix Go 2.4 headset
- [x] Verified battery reporting in the peripherals panel
- [x] Verified dongle disconnect removes the tile and reconnect restores it

## Screenshots
image
image

## Known Limitations / Follow-up
- headset power-off is not yet reflected immediately as `Not connected`; this still needs a better off-state refresh path
- charging detection is not enabled yet because the exact charging-state byte still needs confirmation
- a future headset settings window would be useful for volume-related controls, sound modes, and the auto power-off timeout similar to Armoury Crate


seerge:
@vancinis hello,

That looks interesting, thanks for the research! :) 

Quick question about sound modes and actual settings : do they persist if you completely uninstall Armoury Crate ?

I personally don't have any Asus headsets, but I've got an impression that at least for sound, all processing is done on software level by PC/laptop. This is how Armoury was doing noise cancelling and some other sound effects on older laptops for example (on modern devices it's just handled by realtek drivers)

vancinis:
Hello @seerge, Thanks. 

I have only verified the battery/status path for this device so far. I have not verified yet whether Armoury Crate sound modes persist after uninstalling Armoury Crate, so I do not want to overclaim that part. Based on the USB capture I only confirmed the battery/status request/response path, not any EQ, sound effects, volume processing commands.

So I agree that at least some sound-related features may very well be software-side or driver-side rather than headset-side.

seerge:
@vancinis hello, sorry for a slow response

I have finally found time and now prepare proper support for most modern headsets https://github.com/seerge/g-helper/discussions/5964

The problem wiht GO 2.4 (and ROG Cetra RGB) - is that they are quite old and don't seem to support any fancy EQ / audio stuff on firmware level (i.e. all audio processing is not happening on device itself at all, so there is nothing to toggle or configure)

I have integrated your PR / findings in the build (i.e. battery level / charging status reading), but what would be cool if you can do USBPcap capture of you changing all settings in Armoury for this headset one by one, so I can look deeper and may be add other things that are not software-only

PR #5249, opened 2026-04-15
Design proposal: Mica background

Iemand005:
Here's what GHelper looks like with Mica:

afbeelding

These are the minimal changes to get Mica working.

Dark mode looks great, but light mode needs a little more work to get looking right. I'd also like to make the buttons transparent so the mica effect can show through them.

It might also be a good idea to add a toggle in settings to turn Mica on and off.

Let me know if you'd like me to polish it to get it implemented in release.


seerge:
@Iemand005 hello :) 

The background looks interesting (just at first it didn't work for me, until I realised I had to enable all this "transparency effects" in windows)

But main problem is that all the UI elements I use are effectively custom (mainly overloaded to work for dark theme and have rounded borders). In this setup grey buttons don't really match this transparent blurred background and also corners of the custom elements look jugged due to this.

So honestly I'm not sure if it's worth an effort to try to redo everything for this :)

Thanks

PR #5270, opened 2026-04-19
Aura: Add 50ms delay after init command

LeddaZ:
* Fixes keyboard light sometimes turning off when setting colours in gradient mode

### DIFF
--- a/app/USB/Aura.cs
+++ b/app/USB/Aura.cs
@@ -596,6 +596,7 @@ namespace GHelper.USB
             {
                 initDirect = false;
                 AsusHid.WriteAura(new byte[] { AsusHid.AURA_ID, 0xBC });
+                Task.Delay(50).Wait();
             }
 
             Array.Clear(keyBuf, 0, keyBuf.Length);
--- a/app/USB/Aura.cs
+++ b/app/USB/Aura.cs
@@ -871,7 +871,7 @@ namespace GHelper.USB
                     colors[lightbarOrder[i]] = ColorUtils.GetWeightedAverage(Aura.Color2, Aura.Color1, t);
                 }
 
-                ApplyDirect(colors, true);
+                ApplyDirect(colors);
             }
 
             public static void ApplyGPUColor(int gpuMode = -1)


seerge:
@LeddaZ sorry, I have misread the commit. I thought it adds init and not removes it.

What init is doing is forcing app to run a separate Aura "manual" backlight init
```
AsusHid.WriteAura(new byte[] { AsusHid.AURA_ID, 0xBC });
```

It's needed if laptop is rebooted / was sleeping / etc - and in that case (as I understand) Aura may reset to default and therefore ignore next manual zone color settings.

Why do you think it turns off ? :|

LeddaZ:
Hm, could it be that sending two commands too fast causes it? If I set a breakpoint in Visual Studio right after that manual init is done, and then resume, it always works. But without breakpoint it sometimes stays off.

seerge:
@LeddaZ it could, but the whole purpose of init is that some keyboards just stop responding on custom aura command with zones sometimes, and if you init whole process beforehand it helps :)

Can you try to add a small Thread.Wait after init and check if it helps in your case ?

LeddaZ:
I tried with a 50ms delay after init and gradient works fine, without removing init=true

seerge:
@LeddaZ great, merged then. Thanks

seerge:
@LeddaZ just a quick unrelated question,

I did some deeper reverse-engineering for aura protocol, and now app should be able to detect keyboard backlight type automatically based on aura communication (not by hardcoding model types / names)

Can you quickly check if build from this PR https://github.com/seerge/g-helper/pull/5299 works fine for you, and if possible  also upload log from it ?

Thanks in advance.

LeddaZ:
> @LeddaZ just a quick unrelated question,
> 
> I did some deeper reverse-engineering for aura protocol, and now app should be able to detect keyboard backlight type automatically based on aura communication (not by hardcoding model types / names)
> 
> Can you quickly check if build from this PR #5299 works fine for you, and if possible also upload log from it ?
> 
> Thanks in advance.

I tested the build (I had to edit Aura.cs line 636 to fix a missing */) and everything seems to work fine (it correctly detects I have per-key and lightbar, without logo), here's the log
[log.txt](https://github.com/user-attachments/files/27101013/log.txt)

seerge:
@LeddaZ ok, great! Thanks

I was likely in the middle of pushing more changes.

By the way, does new "gradient" mode fill lightbar in correct order for you ?

LeddaZ:
> @LeddaZ ok, great! Thanks
> 
> I was likely in the middle of pushing more changes.
> 
> By the way, does new "gradient" mode fill lightbar in correct order for you ?

The lightbar doesn't turn on for me, same as ambient/heatmap modes before

seerge:
@LeddaZ ok

But it works for all standard modes, right ?

If so - can you try this build, select a "Test Zone" mode and check if it lights up by any chance ?

[GHelper.zip](https://github.com/user-attachments/files/27102670/GHelper.zip)

LeddaZ:
> @LeddaZ ok
> 
> But it works for all standard modes, right ?
> 
> If so - can you try this build, select a "Test Zone" mode and check if it lights up by any chance ?
> 
> [GHelper.zip](https://github.com/user-attachments/files/27102670/GHelper.zip)

Yes, standard modes still work fine. Zone test still doesn't turn on the lightbar with this build.
[log.txt](https://github.com/user-attachments/files/27103450/log.txt)

seerge:
@LeddaZ hello, one more quick question

Can you check if build in this thread fixes lighbar thing for Ambient/Heatmap/Gradient modes by any chance https://github.com/seerge/g-helper/issues/5579#issuecomment-4720120476 ?

LeddaZ:
Yes, I confirm the lightbar works in all modes on that build

seerge:
@LeddaZ thanks for checking. You have per-key RGB, right ? (not 4-zone one)

LeddaZ:
Yes, per-key

Issue #5383, opened 2026-05-12
Battery Charge Limit - cannot set to 90%


poeggi:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information or no app logs will be provided, my issue will be closed without an answer.

### What's wrong?

I have a GA403WW, when I try to set the Battery Charge limit slider to 90%, it jumps back to 100% immediately.
If however I set it to 80%, it sticks.

### How to reproduce the bug?

Start G-Helper
Open G-Helper
Set Charge-Limit Slider to 90%
See it jump to 100%

### Logs

12 May 2026 08:02:03: Lid Open
12 May 2026 08:02:03: FnLock = 0 : OK
12 May 2026 08:02:03: USB FnLock 19B6|63: 5A-D0-4E-01
12 May 2026 08:02:03: SlashLidCloseAnimation False:5D-D8-00-00-02-A5-80-00-00-00-00-00-00-00-00-00-
12 May 2026 08:02:03: Matrix LidClosed: False
12 May 2026 08:02:03: Matrix Device: \\?\hid#vid_0b05&pid_19b6&col06#7&158463d9&0&0005#{4d1e55b2-f16f-11cf-88cb-001111000030} 128
12 May 2026 08:02:03: SlashSleepInit:5D-D2-02-01-08-A1-00-00-00-00-00-00-00-00-00-00-
12 May 2026 08:02:03: SlashSleep False:5D-D3-03-01-08-A1-00-FF-00-02-FF-FF-00-00-00-00-
12 May 2026 08:02:03: Lid 19B6|63: 5A-BA-C5-C4-00
12 May 2026 08:02:04: SlashEnable False:5D-D8-02-00-01-80-00-00-00-00-00-00-00-00-00-00-
12 May 2026 08:02:04: WMI event 207
12 May 2026 08:02:04: WMI event 123
12 May 2026 08:02:04: WMI event 207
12 May 2026 08:02:04: WMI event 123
12 May 2026 08:02:04: WMI event 207
12 May 2026 08:02:04: WMI event 123
12 May 2026 08:02:04: WMI event 207
12 May 2026 08:02:04: WMI event 123
[...]

PR #5384, opened 2026-05-12
Fix battery slider snapping to unsupported charge limits

poeggi:
On models with restricted charge limits (e.g. GA403), the slider allowed dragging to values like 90% which the backend immediately remapped to 100%, causing a jarring jump on mouse release.

Extract the clamping logic into ClampLimit() and use it to derive the actual supported stop points for the slider at init time, so snapping happens while dragging rather than after letting go.

See #5383

### DIFF
--- a/app/Battery/BatteryControl.cs
+++ b/app/Battery/BatteryControl.cs
@@ -62,18 +62,38 @@ namespace GHelper.Battery
             }
         }
 
+        private static int ClampLimit(int limit)
+        {
+            if (AppConfig.IsChargeLimit6080())
+            {
+                if (limit > 85) return 100;
+                if (limit >= 80) return 80;
+                if (limit < 60) return 60;
+            }
+            return limit;
+        }
+
+        public static int[]? GetSupportedLimits(int step = 1)
+        {
+            if (step <= 0) step = 1;
+            var candidates = Enumerable.Range(0, (100 - 40) / step + 1)
+                .Select(i => 40 + i * step)
+                .ToArray();
+            var stops = candidates
+                .Select(ClampLimit)
+                .Distinct()
+                .OrderBy(x => x)
+                .ToArray();
+            return stops.Length < candidates.Length ? stops : null;
+        }
+
         public static void SetBatteryChargeLimit(int setLimit = -1)
         {
             int limit = setLimit;
             if (limit < 0) limit = AppConfig.Get("charge_limit");
             if (limit < 40 || limit > 100) return;
 
-            if (AppConfig.IsChargeLimit6080())
-            {
-                if (limit > 85) limit = 100;
-                else if (limit >= 80) limit = 80;
-                else if (limit < 60) limit = 60;
-            }
+            limit = ClampLimit(limit);
 
             if (setLimit > 0) SetAsusChargeLimit(limit);
 
--- a/app/Settings.cs
+++ b/app/Settings.cs
@@ -243,6 +243,8 @@ namespace GHelper
             sliderBattery.KeyUp += SliderBattery_KeyUp;
             sliderBattery.ValueChanged += SliderBattery_ValueChanged;
 
+            sliderBattery.Stops = BatteryControl.GetSupportedLimits(sliderBattery.Step);
+
             sensorTimer = new System.Timers.Timer(AppConfig.Get("sensor_timer", 1000));
             sensorTimer.Elapsed += OnTimedEvent;
             sensorTimer.Enabled = sensorsAlways;
--- a/app/UI/Slider.cs
+++ b/app/UI/Slider.cs
@@ -72,14 +72,19 @@ namespace GHelper.UI
                 _step = value;
             }
         }
+
+        public int[]? Stops { get; set; }
+
         private int _value = 50;
         public int Value
         {
             get => _value;
             set
             {
-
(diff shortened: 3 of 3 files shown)


seerge:
@poeggi hello,

Thanks, but I'm not sure if it makes a big difference :) It just prevents you from dragging a circle at all over this range, which can confuse people even more (making them think UI is lagging?)

poeggi:
Obviously I can only speak for myself, i find this solution feels more natural. That said, I though fixing keyboard arrow keys could make sense also ... but trying my own patch - found they were also not working. So I fixed this.

poeggi:
What about this, to make it more clear? 

image

poeggi:
So with this PR we could have two improvements: 1) make slider arrow-key controllable and 2) visually hint towards possible values (on any given platform) => (#2 implementation follwing now)

seerge:
@poeggi thanks, 

With keyboard controls - you can already control it. I agree, it's not super convenient that you need to hold left/right key a bit longer to jump over non-supported range (i.e. you need to let it do 2 steps form 80 to 90 to jump to 100 instead of just clicking a key). From other side it kind of prevents accidental toggles.

Actually whole keyboard control for the slider was initially added for a blind guy.

For highlighting range - I was thinking about this before, I just didn't find a good UI solution that won't look cluttered :|

poeggi:
Your call obviously. All I can do is make proposals.

poeggi:
Or alternative idea - more subtle: 
image

poeggi:
Commiting code for this last (more subtile) proposal - built and tested, works here.

EminGT:
I would like to have this added, I often change between 60 and 80% and keep accidentally setting it to 65/75 for a second. It's no big deal of course but still a quality of life improvement.

seerge:
@EminGT it's a bit different thing :) It just prevents you from dragging tracker at 85-95 range at all

poeggi:
@EminGT , this is what it looks like

https://github.com/user-attachments/assets/e472ef51-4826-41f2-bdfd-47a75e9aeec0

EminGT:
Yeah being able to actually see the dot on the available points makes it so much easier to just click on the right spot on the slider without having to manually move around the slider to "find" 80.

EminGT:
In this video you can see how I'm struggling to go between 60 and 80 back to back, although a bit exaggerated :P

https://github.com/user-attachments/assets/d79a9e78-9f03-4bc1-9df7-e2de0b9eca1c

seerge:
@EminGT if you will click anywhere between 40% and 60% dot will end up on minimum supported value (60%), so it's quite easy :)

poeggi:
@seerge hope you don't mind me disagree as I also felt usability _is_ impaired without this change - that's why I prepared it in the first place. When  I started using G-Helper - which I really love btw! -  I have been convinced for weeks that the charge slider was broken, which is why I wanted to write a bug report earlier (and soon after looking into it found investigating it is not broken, just behaving awkwardly).

Also, having to spam the keyboard or hold button pressed to get past that "dead zone" does not feel like a normal use scenario - rather a quirk (you may get used to).

You can tell, I am still in favor of my own patch ;-) ... but again - waiting for your decision of course.

EminGT:
Yeah even if it snaps to 60% when I click on 40%, it's still just finicky and still a problem when wanting to specifically set it to 70% for example.

poeggi:
Hi @seerge - just wanted to ask if there is anything I can do to help come to a conclusion if this is something to merge or discard?

Here to make any adjustments to the PR if needed.

seerge:
@poeggi hi, sorry for a slow response :) I didn't had time to settle in my head yet.

In reality I don't see it as very big UI problem in a first place. 

Honestly, i don't understand why Asus has started restricting charge limits at all in newer models, as on G14 2022 (the one i used to develop app initially) it was possible to set any limit, even 95%, that's from where whole UX/UI solution comes from. And I even had idea to add actual battery charge to a same bar (but didn't add it yet) :)

xxGonzalesxx:
Hey guys, nice to see you. I'm starting to learn C# after Java, as I've found some drawbacks. Now I'm trying to write Stealer.

seerge:
@poeggi @EminGT hey guys,

I was preparing some minor UI tweaks in the app (most of them already live in 0.256) to match native winUI3 theme (without carrying huge SDK inside).

And as one of the changes, is animation of the slider-head to make circle bigger/smaller on hover/drag (same as native winUI element behaves)

And since it already adds some mouse in/out animations I have decided to try @poeggi 's idea (thanks) with ticks, but show it only for mouse over state (so we keep actual app UI clean)

Ticks woud appear only for battery slider, and only for 60-80 models (where it's flexible - i would keep things simple)

This build does it

https://github.com/seerge/g-helper/pull/5558

What do you think ?

poeggi:
Hi @seerge ... looks like an improvement - although my personal taste the tick markers are a little to light and I definitely like the "snap-to-value" concept. As always - YMMV

seerge:
@poeggi you mean make ticks even whiter than they are now (on dark background ?)

As for snapping, I was changing sliders to behave as close to native WinUI for win11 as much as possible, and here it's smooth dragging

poeggi:
I think the stroke width of the ticks should match that of the fonts, also the ticks are currently not visible when the slider gains focus via tabbing in.

PR #5444, opened 2026-05-24
Sync Fans

HamzaY2:
# Sync Fan Speeds

This PR adds a new **Sync Fan Speeds** option for fan control.

The idea is simple: when enabled, all active fans follow one shared curve based on the hottest component at that moment. This makes the cooling behavior a bit more practical for real-world workloads where the CPU and GPU do not always heat up at the same time.

## Why

Right now, each fan follows its own curve and only reacts to its own sensor. That works in many cases, but it can fall short when the GPU heats up before the CPU.

For example, the GPU might already be at 55 °C while the GPU fan is still following the 55 °C point on its own curve. In some setups, that means it stays quieter than expected, even though the shared CPU curve would have ramped the fans up more aggressively and helped cool the GPU sooner.

Until now, the workaround was to manually copy the CPU curve to the GPU and Mid fan charts, then keep them in sync by hand. That gets annoying quickly, especially once you start tweaking profiles.

This PR adds a toggle to handle that automatically.

## What changed

A new **Sync Fan Speeds** checkbox was added above **Apply Fan Curve** on the Fans & Power page.

When both **Apply Fan Curve** and **Sync Fan Speeds** are enabled:

- CPU and GPU temperatures are checked every 2 seconds.
- The higher temperature is used as the source value.
- That temperature is looked up on the CPU fan curve.
- The resulting fan percentage is applied to every active fan.
- GPU and Mid fan charts are shown as disabled and flattened to the current synced percentage.
- The CPU chart stays editable and becomes the shared curve source.

So the CPU curve effectively becomes the universal fan curve while Sync is enabled.

When Sync is turned off, the normal per-fan behavior comes back. Saved CPU, GPU, and Mid curves are reloaded and pushed back to the hardware.

This does not overwrite any stored fan curves.

## Notes

- The setting is saved per profile, just like the existing fan settings.
- Sync depends on **Apply Fan Curve** being enabled.
- Turning **Apply Fan Curve** off will also turn Sync off.
- Turning **Apply Fan Curve** back on will not automatically re-enable Sync.
- Hardware writes are only sent when the target percentage changes.
- The patch is intentionally narrow and does not touch power, RGB, or unrelated profile logic.

## Files touched

- `app/Fans.cs`
  - sync timer
  - hottest-temperature lookup
  - start/stop lifecycle
  - per-profile persistence

- `app/Fans.Designer.cs`
  - new toggle UI

- `app/Properties/Strings.resx`
- `app/Properties/Strings.tr.resx`
- generated designer files
  - English: `Sync Fan Speeds`
  - Turkish: `Fan Hızlarını Eşitle`
  
  Co-Authored: @HamzaYslmn


PurpleSky-8Bit:
I’ve always wanted this feature... since stressing a single fan isn’t ideal... especially in laptops with a shared cooling system. When I use the laptop in Eco mode only the CPU fan works and the GPU fan totally stops working,

HamzaY2:
if you want deep search use [this tool](https://www.asuswebstorage.com/navigate/a/#/s/A8E59DDAEDD745FDB00B0D12C89D6D274), asus original log dump tool

HamzaY2:
@seerge

PR #5448, opened 2026-05-24
Read-only fan & CPU-temp support for Snapdragon Zenbook A16 (UX3607OA, ARM64)

agentseo:
### What this adds

Read-only monitoring (fans + CPU temperature) for the **ASUS Zenbook A16 UX3607OA** — a Snapdragon X2 (ARM64) Copilot+ laptop. On stock g-helper this model shows **no CPU temperature** and only **one mis-decoded fan**. All changes are gated behind `AppConfig.ContainsModel("UX3607")`, so **the x86 code path is completely untouched**.

This is **monitoring only** — see "Why no fan control" below.

### Changes

**`app/AsusACPI.cs` — `GetFan()`**

1. **Packed RPM decode.** On this firmware `CPU_Fan` (0x00110013) returns a packed value: a status/mode field in the high word and the speed (hundreds of RPM) in the low word. Real fan readings never exceed `0xFFFF`, so values above that are masked to the low word. Gated by magnitude, not model, so it's a no-op on normal hardware.
2. **Second fan tach.** This model has a real second fan, but the stock `Mid_Fan` ID (0x00110031) returns 0 for it. The working tach lives at a non-standard ID `0x00110035`. A model-gated field `_midFanReadId` routes the "Mid" slot to that ID on UX3607 only (`Mid_Fan` everywhere else).

**`app/HardwareControl.cs` — CPU temperature**

3. **WMI thermal-zone fallback.** ACPI `Temp_CPU` (0x00120094) returns 0 on this model (ASUS didn't wire the ATK temp endpoint on this ARM SKU). When the existing temp read is `<= 0`, fall back to `MSAcpi_ThermalZoneTemperature` over WMI.
4. **Prefer TZ33 on UX3607.** Of the OS-registered thermal zones, the `QCOM*` zones report static, baked-in values (~41 °C, frozen, never move under load). `TZ33` is the only zone that tracks real SoC temperature. Verified by load test: ~34 °C idle → ~43 °C under GPU load → ~48–56 °C under CPU load, cross-checked against HWiNFO. Without this, max-of-zones would pin to the fake QCOM reading. Gated by `_preferTZ33` (UX3607 only); other models keep max-of-zones behaviour.

### Why no fan _control_

Fan **control** is impossible on this firmware: the DEVS fan-write endpoints present on x86 ASUS laptops (`DevsCPUFan` 0x00110022, `DevsCPUFanCurve` 0x00110024, `FanHysteresis` 0x00110034) are **absent** in this model's firmware (confirmed via the built-in ACPI scan). So this PR deliberately does not attempt any fan write — it only reads.

### Scope & safety

- Every change is behind a model check (`ContainsModel("UX3607")`) or a value-range check; **no behaviour change on any existing device**.
- Read-only; no new ACPI writes.
- The WMI temp read uses the elevated context g-helper already runs in (scheduled task), so no new privilege requirement in normal use.

### Testing

Verified on the actual UX3607OA hardware:

- Both fans read correct RPM and rise synchronously under load.
- CPU temperature tracks real load (idle / GPU stress / CPU stress), cross-validated against HWiNFO64.

### Notes for the maintainer

This is a niche ARM SKU and I understand if Snapdragon support is out of scope for the project. The change is intentionally small, self-contained and opt-in so it carries near-zero risk for existing users. Happy to adjust the model gate, naming, or split the commits if you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/AsusACPI.cs
+++ b/app/AsusACPI.cs
@@ -505,11 +505,19 @@ public class AsusACPI
                 fan = Program.acpi.DeviceGet(GPU_Fan);
                 break;
             case AsusFan.Mid:
-                fan = Program.acpi.DeviceGet(Mid_Fan);
+                fan = Program.acpi.DeviceGet(_midFanReadId);
                 break;
             default:
                 fan = Program.acpi.DeviceGet(CPU_Fan);
                 break;
+        }
+
+        // Some firmware (e.g. Snapdragon Zenbook UX3607) packs a status/mode field in
+        // the high word and the fan speed (hundreds of RPM) in the low word. Real fan
+        // readings never exceed 0xFFFF, so an out-of-range value means it's packed.
+        if (fan > 0xFFFF)
+        {
+            fan = fan & 0xFFFF;
         }
 
         if (fan < 0)
@@ -520,6 +528,11 @@ public class AsusACPI
 
         return fan;
     }
+
+    // UX3607 (Snapdragon) has a real second fan whose tach lives at a non-standard
+    // ID; the stock Mid_Fan (0x00110031) returns nothing on this model. Route the
+    // Mid slot to the working ID so the second fan's RPM is displayed.
+    static readonly uint _midFanReadId = AppConfig.ContainsModel("UX3607") ? 0x00110035u : Mid_Fan;
 
     public bool IsMidFanSupported()
     {
--- a/app/HardwareControl.cs
+++ b/app/HardwareControl.cs
@@ -449,6 +449,17 @@ public static class HardwareControl
             //Debug.WriteLine("Failed reading CPU temp :" + ex.Message);
         }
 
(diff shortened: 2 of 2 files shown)


mirh:
FWIW this is how X1 did it
https://github.com/icecream95/x1e-ec-tool
https://git.kernel.org/pub/scm/linux/kernel/git/qcom/linux.git/tree/drivers/platform/arm64/qcom-hamoa-ec.c?h=arm64-for-7.3

Could you tell whether the EC is by ITE [or](https://lore.kernel.org/all/ad3485df-5570-4631-9d6e-2f0b90a47235@oss.qualcomm.com/) NPCX?
EDIT: [official driver](https://lore.kernel.org/all/20260922-topic-asus_ec-v2-2-0bd923d1b853@oss.qualcomm.com/)

mirh:
Needs rebasing.

PR #5471, opened 2026-05-28
Standalone overlay

seerge:
Runs Overlay in a separate mini-process placed into `C:\Program Files\GHelper` folder (windows requiremenets), which allows overlay to stay on top of ALL full-screen apps.

# [GHelper.zip](https://github.com/user-attachments/files/29467343/GHelper.zip)
v0.261


ken-kuro:
Do we have a timeline to release this :D ?

seerge:
@ken-kuro for this one - not sure, it has all overlay logic moved into mini app, that makes things quite complicated to achieve one benefit to draw above full-screen games 

But this one https://github.com/seerge/g-helper/pull/5735 has higher chances (as here app runs a mini-app just to display data, all logic / overlay monitoring happens in GHelper.exe)

I'm just not sure if all this workarounds worth an effort in a first place

PR #5476, opened 2026-05-29
Design changes

seerge:
Subtle borders around main controls + light gradient higlights

Screenshot 2026-06-06 222015


RedSTARO:
Hi, I'd like to share some feedback on the recent UI update. The new interface feels noticeably more skeuomorphic than before, while I preferred the flat design of the previous version. Would it be possible to add a setting that lets users switch back to the older UI? Having that as an option would be much appreciated. Thanks!

seerge:
@RedSTARO hello, it's actually just a tiny gradient borders around main elements :) Nothing more.

You can disable them still tho
https://github.com/seerge/g-helper/wiki/Power-user-settings#flat-ui-theme

RedSTARO:
Thank you so much!2026年6月27日 17:12，Serge ***@***.***>写道：seerge left a comment (seerge/g-helper#5476)
@RedSTARO hello, it's actually just a tiny gradient borders around main elements :) Nothing more.
You can disable them still tho
https://github.com/seerge/g-helper/wiki/Power-user-settings#flat-ui-theme

PR #5558, opened 2026-06-12
Slider tweaks

seerge:
[GHelper.zip](https://github.com/user-attachments/files/28883826/GHelper.zip)


EminGT:
I like it, definitely an improvement :)

Issue #5497, opened 2026-06-03
Ghelper new monitoring rivals with RivaTuner


ghost:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information will be provided, my issue will be closed without an answer.

### Is your feature request related to a problem? Please describe

Hello dear Serge, Im a long time user and been donating to you a couple of times.

I own Asus scar 16 275HX 5090.
Your new feature on Ghelper, called "Hardware overlay" is incredibly good. It can rival with RivaTune overlay used with msi afterurner. But yours wins. Let me explain why.

I'm a heavy user of Nvidia Profile Inspector and for some games I use Smooth Motion , to enable frame generation on unsupported games like Baldur's Gate 3 and Red Dead Redemption 2. Enabling Smooth Motion doesnt work with RivaTuner, only when RivaTuner is OFF. 

I found that your monitoring overlay doesn't conflict with Smooth Motion. There is one feature that could be added : editing the overlay. Basically allow us to users, edit the name for ex I like to set the name of my GPU (5090) CPU name. 
Here is the example of my RivaTuner, Im sure something like that could be done with Ghelper overlay.

<img width="1406" height="585" alt="Image" src="https://github.com/user-attachments/assets/fa5fe2ba-86a9-4ad5-88fa-1b907dc6261c" />

 This would from my guess make a totally new software if you decide to separate it for non-Asus users, because this will replace RivaTuner on a worldwide scale... imagine the potential!  Everyone is gonna use your overlay.

Would it be possible on ghelper?

Thank you

### Describe the solution you'd like

To add option to edit overlay names to match GPU/CPU naming, size of FPS numbers... the ability to remove rpms or anything that user wants to remove.

### Device and Model

G635LX

seerge:
@xjogass hello, 

Thanks for a nice feedback.

But what is the practical sense to constantly show you GPU name ? You already know it better than anyone.

I have absolutely intentionally carefully designed layout to be minimal and show only things that matter (within 3 possible display modes), I don't want to overload it even more with absolutely static data.

Thanks

seerge:
[...]

PR #5562, opened 2026-06-12
Overlay complete mode

seerge:
image

`CTRL + SHIFT + ALT + Mouse click` to toggle between overlay modes


abdellatif-laghjaj:
@seerge  Why we even need the name of the gpu and cpu in the overlay ?? i guess it's useless and just take space

seerge:
@abdellatif-laghjaj i also don't understand that at all (actually RAM usage is also not informative due to a way modern windows manages it), but apparently some like to brag with their hardware ? https://github.com/seerge/g-helper/issues/5497

abdellatif-laghjaj:
> @abdellatif-laghjaj i also don't understand that at all (actually RAM usage is also not informative due to a way modern windows manages it), but apparently some like to brag with their hardware ? #5497

At least make make customizble ? i mean users can check what info to be displayed in the oevrlay, and this way evryone will be happy, and its actaully a good UX

seerge:
@abdellatif-laghjaj yeah, i have made it optional, under param
```
"overlay_names": 1,
```

PR #5606, opened 2026-06-19
This PR implements automatic performance profile switching based on the active foreground window.

TUSHAR91316:
### Implementation Details
1. **Process Monitoring & Event Hooks (`AppProfileWatcher.cs`)**
   * Uses Win32 `SetWinEventHook` with `EVENT_SYSTEM_FOREGROUND` to detect active window changes. There are no continuous loops or periodic polling threads; code is triggered only when the active window changes.
   * To prevent heavy `.NET` allocations, process name resolution avoids `Process.GetProcessById()` (which does deep queries of threads/modules). Instead, it uses direct Win32 calls (`OpenProcess` with `PROCESS_QUERY_LIMITED_INFORMATION` + `QueryFullProcessImageName`), which run in microseconds with zero GC allocations.
   * Shell and system processes (`explorer`, `ghelper`, `taskmgr`, etc.) are ignored so G-Helper retains the process name of the actual user application.
2. **Mappings & Config Cache**
   * Mappings are loaded once on startup into a private `_appProfiles` dictionary cache, ensuring O(1) lookups without string parsing on window switches. Mappings are stored in G-Helper's existing `config.json` via the `"app_profiles"` key (stored as `process:mode` csv pairs).
3. **Edge Case Safety**
   * **Cooldown Timer**: Uses a `1000ms` cooldown timer. If a user quickly switches between windows or Alt-Tabs, G-Helper waits for the focus to settle before invoking driver mode changes.
   * **Manual Syncing**: If a user manually changes the mode (via keyboard hotkeys or G-Helper UI) while in a mapped application, that mode is treated as the new manual default fallback, and G-Helper won't try to override it until they switch applications again.
   * **Fallback**: Gracefully restores the manual default profile when focusing back on any unmapped app (like web browsers or File Explorer).
4. **Zero UI Footprint Context Menu (`Settings.cs`)**
   * Adds an `"Auto Profile: [AppName]"` submenu right before the `"Quit"` item in G-Helper's tray icon context menu.
   * Rebuilds dynamically on the context menu `Opening` event, ensuring the detected foreground process is fresh right before display.
   * When clicked, G-Helper dynamically detects the last active user process name and populates the sub-menu with available profiles (Silent, Balanced, Turbo, and custom profiles) with a checkmark on any existing link.
   * If a link exists, a `"Remove Profile Link"` item is displayed to clear the mapping.
---
### Future Roadmap / Planned Enhancements
To make this system even more robust while preserving G-Helper's core design principles, future iterations will look into:
* **G-Helper App-Wise Profile Manager**: Introduce a simple, lightweight grid or list within G-Helper's settings window to view, add, or clear app profile associations manually (complementing the tray context menu).
* **Armoury Crate Style Profile Tuning**: Expand the watch actions to map individual custom CPU power limits, fan curves, or GPU modes (Eco, Standard, Optimized) directly to specific application focus events (scenario profile building).

Screenshot 2026-06-19 170038


TUSHAR91316:
@seerge check this

PR #5619, opened 2026-06-22
AMD boost profile

Loganser:
@seerge An implementation of https://github.com/seerge/g-helper/discussions/4711
Tested on ROG Flow Z13 2025.

image

### DIFF
--- a/app/Fans.Designer.cs
+++ b/app/Fans.Designer.cs
@@ -31,6 +31,7 @@ namespace GHelper
         /// </summary>
         private void InitializeComponent()
         {
+            components = new System.ComponentModel.Container();
             ChartArea chartArea1 = new ChartArea();
             Title title1 = new Title();
             ChartArea chartArea2 = new ChartArea();
@@ -97,6 +98,14 @@ namespace GHelper
             panelTitleTemp = new Panel();
             pictureTemp = new PictureBox();
             labelTempLimit = new Label();
+            panelTitleAmdProfile = new Panel();
+            pictureAmdProfile = new PictureBox();
+            labelTitleAmdProfile = new Label();
+            pictureAmdProfileHelp = new PictureBox();
+            panelAmdProfile = new Panel();
+            comboAmdProfile = new RComboBox();
+            panelAmdProfileDivider = new Panel();
+            toolTip = new ToolTip(components);
             panelDownload = new Panel();
             buttonDownload = new RButton();
             panelPawnIO = new Panel();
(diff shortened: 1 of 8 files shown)


seerge:
@Loganser hello, thanks but what is the practical purpose of it ? :) 

In my experience (with G14 2022 / 2025 and ROG Ally, all have AMD CPU) - AMD follows windows power mode guidance quite well, so there is no real reason to override it via RyzenSMU on top of it.

I'm absolutely intentionally trying to keep app UI as simple as possible (and still accommodate all Asus proprietary stuff + use overrides only when they can't be achieved otherwise, like undervolting) :)

Loganser:
On APU it's mostly VRAM boost behavior and power consumption for not demanding tasks, the difference is quite big sometimes. For CPU - not sure, cannot test it, but it also should have simillar effect.
Also without this option it's impossible to achieve truly identical chip behavior pluggud in / on battery.

I understand it's not a necessarry option for such a minimalistic app but still I would really appreciate having it, at least by enabling it with additional config flag.
Anyway, decision is up to you. BTW thank you for your work, this app is amazing!

seerge:
@Loganser can you post any real life benchmark with measurements showing that for different "boost" setting ?

After that, try to turn it off completely, reboot (to reset to default) and just change power mode instead.

Loganser:
## In game:
Balanced+BoostPerformance:
Balanced+BoostPerformance

Balanced+BoostSaving:
Balanced+BoostSaving

BestPerformance+BoostPerformance:
HighPerformance+BoostPerformance

BestPerformance+BoostSaving:
HighPerformance+BoostSaving

BestPowerEfficiency+BoostDefault:
BestPowerEfficiency_BoostDefault
-

## 4K youtube video:

HighPerformance+BoostPerformance:
Video_HighPerformance_BoostPerformance

HighPerformance+BoostSaving:
Video_HighPerformance_BoostSaving

Balanced+BoostPerformance:
Video_Balanced_BoostPerformance

Balanced+BoostSaving:
Video_Balanced_BoostSaving

BestPowerEfficiency_BoostDefault:
Video_BestPowerEfficiency_BoostDefault
-

Summing it up, boost mode is completely independent from windows power mode (and any power plan settings), at least on Flow Z13 2025.

seerge:
@Loganser what game is that and it seem to be FPS limited ?

I tried this "boost" toggle on my Ryzen AI 9 HX (G14 2025) in some CPU heavy games, that use all CPU power when they are allowed to and at the same time run just with same FPS when power is limited manually like Civ 7

And both "boost" states don't make any difference. So it doesn't seem to tame CPU at all.

PR #5712, opened 2026-07-15
Claude/jolly newton az7frl

Ayushprtp:


seerge:
It already exists for ages  https://github.com/seerge/g-helper-experimental-release

PR #5713, opened 2026-07-16
Fix missing GPU temperature on AMD iGPU-only models (Strix Halo / GZ302)

sumit-m:
On ROG Flow Z13 2025 (GZ302EA, Ryzen AI Max "Strix Halo", Radeon 8060S) G-Helper shows GPU fan RPM but no GPU temperature.

Two causes:

1. `AmdGpuControl`'s constructor returns early when `AppConfig.NoGpu()` is true, and GZ302 is on the NoGpu list — so ADL is never initialized and `GetiGpuSensors()` always returns nulls on this model. Changed the guard to still initialize ADL when `AppConfig.IsAMDiGPU()` is true, since those models need it for iGPU sensors.

2. `GetiGpuSensors()` reads `PMLOG_TEMPERATURE_EDGE`, which Strix Halo doesn't report (sensor unsupported in `ADL2_New_QueryPMLogData_Get` output). It exposes `PMLOG_TEMPERATURE_GFX` (28) and `PMLOG_TEMPERATURE_SOC` (29) instead. Added a fallback chain EDGE → GFX → SOC; devices that report EDGE (e.g. ROG Ally) are unaffected.

Verified on GZ302EA (Windows 11, Adrenalin 32.0.23033.5002): temperature now shows next to GPU fan speed and tracks load correctly (48°C idle → ~90°C under sustained GPU load), matching values read directly from the driver's PMLog interface. GPU usage and power readings on iGPU-only models also benefit from fix 1, as they go through the same `GetiGpuSensors()` path.

### DIFF
--- a/app/Gpu/AMD/AmdGpuControl.cs
+++ b/app/Gpu/AMD/AmdGpuControl.cs
@@ -69,7 +69,8 @@ public class AmdGpuControl : IGpuControl
 
     public AmdGpuControl()
     {
-        if (AppConfig.NoGpu() || !Adl2.Load()) return;
+        // AMD iGPU-only models (e.g. GZ302) still need ADL for iGPU sensors
+        if ((AppConfig.NoGpu() && !AppConfig.IsAMDiGPU()) || !Adl2.Load()) return;
 
         try
         {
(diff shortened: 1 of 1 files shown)


seerge:
@sumit-m hello, thanks for spotting

I'm actually preparing a more general iGPU overlay support for all devices where it's reasonable https://github.com/seerge/g-helper/pull/5719

So I have included your fallback to PMLOG_TEMPERATURE_GFX  / PMLOG_TEMPERATURE_SOC  here too.

Can you check build from that branch and say if it shows readings correctly, also on my G14 with AMD iGPU FPS didn't seem to read in all cases when apps were running on iGPU, now it should show that too (not sure if it applies to 8060S)

sumit-m:
Superseded by #5719, which generalizes iGPU overlay support and includes this GFX/SOC temperature fallback. Verified working on my GZ302EA. Closing in favor of that. Thanks @seerge!

Issue #4411, opened 2025-08-03
option to have gpu and cpu fans and use higher temperature between cpu and gpu


ouya-woelders:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information will be provided, my issue will be closed without an answer.

### Is your feature request related to a problem? Please describe

no, well partially. laptops that have 2 fans have one heatsink with heatpipes running across cpu and gpu, when g-helper only runs one fan at full speed and the other at lets say 50% speed under 100% gpu load or 100% cpu load, cooling is less good than with both fans at full speed. for example in eco gpu mode, the cpu runs cooler than in standard mode, because in eco the fans are both at the same speed.

### Describe the solution you'd like

option that allows for both gpu and cpu fans to use one temperature reading for fan speed, this temperature reading is the max temp between cpu and gpu

### Describe alternatives you've considered

eco mode behaves like this, but gpu is not enabled

### Device and Model

proart px13 hn7306wv

ouya-woelders:
or in other words have all fans act on the hottest part instead of having fans act separately for each component

seerge:
@ouya-woelders hello,

Please make yourself familiar with [FAQ](https://github.com/seerge/g-helper/wiki/FAQ#how-does-g-helper-control-my-fan-speeds)

> [### How does G-helper control my fan speeds?](https://github.com/seerge/g-helper/wiki/FAQ#how-does-g-helper-control-my-fan-speeds)
> **It doesn't and can't control your fans**. Firmware / BIOS controls them in real-time. Armoury also doesn't control fans in real time anyhow.

> What G-helper can do - is (optionally) set a custom fan profile to the current performance mode via the same endpoint Armoury uses in Manual mode.

> How it will be interpreted - is still up to the firmware. If you don't like how firmware controls fans, you can try [Experimental build with a manual fan control](https://github.com/seerge/g-helper/discussions/2272)

If you want to have completely manual fan control over fans you you can try [Experimental build with a manual fan control](https://github.com/seerge/g-helper/discussions/2272)

And set ```"fan_shift" : 50,``` as explained in the post

ouya-woelders:
[...]

PR #5727, opened 2026-07-17
Optional fan sync to hottest sensor for shared heatsinks

Lafnaps:
Hi, this adds an opt-in workaround for something that's been bugging me on my M16 (GU604VY) and was asked before in #4411.

With a custom fan curve applied, EC ties each fan strictly to its own sensor. On a shared heatsink this gives two effects:

- GPU fan stays at ~2000 rpm while the CPU sits at 95C, even though both fans cool the same heatpipes
- when the dGPU keeps dozing off and waking up, its fan starts and stops every couple of minutes (EC seems to jump between the GPU sensor and the CPU fallback), which is quite audible in a quiet room

Stock profiles don't do this, fans there follow the hottest component more or less.

What the patch does: a new "Sync Fans to Hottest Sensor" checkbox in Fans + Power. While it's enabled (and only while custom curves are actually applied), a 3s timer takes max(cpu, gpu) temp and re-applies each fan's own curve with the speed points raised to at least what that curve prescribes at that temp. So it's still the usual fan curve endpoint and EC still does all realtime control - fans just can't drop below the "hottest sensor" floor. Curves are only rewritten when the floor actually changes (2C step), so at steady temps there is no extra ACPI traffic at all.

Sync is stopped around calibration, factory reset and sleep/shutdown resets so it never fights them, and it stays inactive when curves are off or were rejected by firmware.

Tested on GU604VY (3 fans incl. mid): CPU load at 95C takes the GPU fan from ~2100 to ~3900 rpm; calibration, mode switching and plug/unplug behave as before with the option off. I understand fan logic is a sensitive area - the feature is fully opt-in and defaults to off.

### DIFF
--- a/app/Fan/FanMaxTempControl.cs
+++ b/app/Fan/FanMaxTempControl.cs
@@ -1,0 +1,111 @@
+namespace GHelper.Fan
+{
+    // EC firmware runs each fan off its own sensor (CPU fan - CPU temp, GPU fan - GPU temp).
+    // Since the heatsink is shared, this periodically raises the floor of every custom curve
+    // to the speed its own curve prescribes at the hottest sensor, so no fan idles while
+    // another component is hot. EC keeps reacting to its own sensor above the floor.
+    //
+    // Lifecycle: started by ModeControl.AutoFans after custom curves are successfully applied,
+    // stopped by AutoFans otherwise and by every path that resets EC curves behind its back
+    // (calibration, factory reset, sleep/shutdown reset).
+    public static class FanMaxTempControl
+    {
+        const int TICK_MS = 3000;
+        const int TEMP_STEP = 2;
+        const int NO_TEMP = -1000;
+
+        static readonly System.Timers.Timer timer = new(TICK_MS);
+        static readonly object syncLock = new();
+
+        static int lastTemp = NO_TEMP;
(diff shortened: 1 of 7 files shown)


seerge:
@Lafnaps hello, it's already possible with https://github.com/seerge/g-helper-experimental-release that has a proper manual fan control

Adjusting fan curve won't achieve this anyway, as it's still up to a free firmware interpretation :)

Lafnaps:
Thanks for the reply! I did check out the experimental build before going this route. My main concern with proper manual control is the failure mode: the app is now the thing driving the fans, so if it crashes or gets killed, the fan is left stuck at whatever was last set, with nothing watching the temps anymore.

That's exactly what this PR avoids - the firmware stays in charge the whole time. All it does is raise the floor of the curve that gets written to the EC, so the worst case (app dies) is just a valid curve that the EC keeps executing on its own. Fail-safe by design.

On the "free firmware interpretation" point - fair, but I did verify it on real hardware (GU604V): with a hot CPU and an idle dGPU, writing the synced curve took the GPU fan from ~2100 to ~3900 rpm, so both fans end up spinning at the speed the hottest sensor calls for. Which is exactly what #4411 was about.

PR #5719, opened 2026-07-16
iGPU sensors for APU-only models

seerge:
image

 iGPU overlay support for AMD APU only GPUs + fixed FPS readings for apps that run on iGPU / Eco mode

[GHelper.zip](https://github.com/user-attachments/files/30166455/GHelper.zip)

PR #5740, opened 2026-07-19
Fix Strix Halo overlay sensors when ADL PMLog is unavailable

jkwind02-lang:
## Summary  This is an incremental fix for #5719 on a ROG Flow Z13 GZ302EA with Radeon 8060S (Strix Halo).  On this device ADL enumerates five duplicate adapters with `VendorID = -1002`. `ADL2_Adapter_ASICFamilyType_Get` returns `-8` for every index, so the integrated adapter is never selected. Even after selecting it, `ADL2_New_QueryPMLogData_Get` returns `-8` for all five indices; the legacy Overdrive temperature and performance APIs return `-5`.  The patch:  - accepts the signed AMD vendor ID and uses the configured AMD-iGPU fallback when ASIC family classification is unavailable; - keeps ADL PMLog as the preferred source; - uses AMD ADLX as the primary fallback for temperature, utilization, current VRAM and GPU power; - keeps Windows `D3DKMT_ADAPTER_PERFDATA`, matching-LUID `GPU Engine` counters and video-memory information as final fallbacks; - stops deriving iGPU power by subtracting per-core Energy Meter counters from the CPU package counter; - uses the sum of the per-core RAPL Energy Meter counters for AMD Software-compatible CPU power, while retaining package power only as a generic fallback on other systems.  ## Verification  - `dotnet build app/GHelper.csproj -c Release --no-restore`: succeeded with 0 warnings and 0 errors. - `dotnet publish app/GHelper.csproj -c Release -r win-x64 --self-contained false -p:PublishSingleFile=true --no-restore`: succeeded. - Runtime on GZ302EA / Radeon 8060S / driver 32.0.23033.5002:   - ADLX temperature and utilization match AMD Software sampling much more closely than the Windows busiest-engine fallback;   - ADLX VRAM current usage is returned correctly on the UMA adapter;   - under the same active workload, ADLX `GPUPower` reported 50-62 W while the old package-minus-core estimate was around 10-20 W;   - ADLX `GPUTotalBoardPower` is unsupported/zero on this APU, so `GPUPower` is used;   - same-screen comparison: AMD Software CPU power was 25.9 W and the 16 per-core RAPL counters summed to about 25.9 W, while `RAPL_Package0_PKG` was 48-54 W because it includes shared APU/GPU domains.  The ADLX path uses `ADLXInitializeWithIncompatibleDriver`, because regular `ADLXInitialize` returns an empty GPU list with this OEM driver.

### DIFF
--- a/app/Gpu/AMD/AdlxGpuSensor.cs
+++ b/app/Gpu/AMD/AdlxGpuSensor.cs
@@ -1,0 +1,133 @@
+using System.Runtime.InteropServices;
+
+namespace GHelper.Gpu.AMD;
+
+internal static class AdlxGpuSensor
+{
+    private const ulong AdlxVersion = ((ulong)1 << 48) | ((ulong)5 << 32) | 124;
+    private const int CacheMs = 500;
+    // Function indexes from the backward-compatible ADLX 1.5 C interface vtables.
+    private const int ReleaseIndex = 1;
+    private const int SystemGetGpusIndex = 1;
+    private const int SystemGetPerformanceIndex = 9;
+    private const int ListBeginIndex = 5;
+    private const int GpuListAtIndex = 11;
+    private const int PerformanceGetCurrentGpuIndex = 18;
+    private const int MetricsUsageIndex = 4;
+    private const int MetricsTemperatureIndex = 7;
+    private const int MetricsVramIndex = 12;
+
+    private static readonly object Sync = new();
+    private static bool _initialized;
+    private static bool _initializationAttempted;
+    private static nint _system;
+    private static nint _gpu;
+    private static nint _performance;
+    private static long _lastRead = -CacheMs;
+    private static (int? temp, int? use, int? vramUsedMb) _metrics;
+
+    [DllImport("amdadlx64.dll", CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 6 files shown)


jkwind02-lang:
Added power-metric alignment in 5b61fa3c after testing against the AMD driver metrics on the affected GZ302EA. The iGPU power now comes directly from ADLX `GPUPower` (50-62 W in the captured workload) instead of the package-minus-core estimate, and CPU power keeps the `RAPL_Package0_PKG` reading intact. Build and single-file publish both pass.

jkwind02-lang:
Follow-up CPU power correction in cc693561 after a same-screen comparison with AMD Software on the affected GZ302EA:

- AMD Software CPU power: 25.9 W
- Sum of the 16 `RAPL_Package0_Core*_CORE` counters at the same workload: about 25.9 W
- `RAPL_Package0_PKG`: 48-54 W, tracking the shared APU/GPU domains rather than AMD Software's CPU-only metric

The overlay now reads CPU power from the per-core RAPL sum and GPU power independently from ADLX. It no longer uses package power as CPU power and no longer derives GPU power by subtraction. Build and single-file publish pass with 0 warnings/errors.

PR #5767, opened 2026-07-23
Fix driver check for Realtek LAN Driver

LeddaZ:
Realtek encodes the chip family in the version major for LAN drivers (e.g. 1168.x = RTL8168, 1125.x = RTL8125). Asus lists multiple branches under the same title, so keep the newest of each family instead of only the first, and then hide drivers for Realtek LAN cards that aren't installed.

### DIFF
--- a/app/UpdatesController.cs
+++ b/app/UpdatesController.cs
@@ -1,4 +1,3 @@
-using GHelper.Helpers;
 using Microsoft.Win32;
 using System.Management;
 using System.Net;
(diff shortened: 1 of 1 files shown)


seerge:
@LeddaZ hello, thanks for a discovery.

PR seem to also compete with some other tweaks I did already (by hiding duplicates)

Can you just check if this build with Realtek version specific fix shows correct drivers only for you ?

[GHelper.zip](https://github.com/user-attachments/files/30352395/GHelper.zip)

LeddaZ:
I can confirm it shows only the correct drivers

seerge:
@LeddaZ great! It's live in 0.268

PR #5779, opened 2026-07-25
fixed OLED not being detected for ASUS Vivobook 14 Flip TP3407

atom501:
Flicker-free dimming option was not showing up for my laptop, ASUS Vivobook 14 Flip TP3407SA. I found that the function IsOLED() was returning false.

One solution was to add my laptop's model to the list of ContainsModel() checks. But I found that my laptop has different registry entries for OLED. The closest thing to "SOFTWARE\ASUS\OLEDCare" was:

oled care

But the OLED related values are present here:

oled values

Adding an additional read for OLEDPanel value fixed the issue. IsOLED() returned true and Flicker-free dimming bar showed up when running the compiled application.

### DIFF
--- a/app/AppConfig.cs
+++ b/app/AppConfig.cs
@@ -477,7 +477,12 @@ public static class AppConfig
         try
         {
             var key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\ASUS\OLEDCare");
-            return key is not null && Convert.ToInt32(key.GetValue("EnablePixelRefresh", 0)) != 0;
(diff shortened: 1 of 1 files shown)


seerge:
@atom501 hello, thanks, for in your case it's easier to add model by SKU (TP3407) to the OLED list (which i just did) instead of pulling one more registry key :)

Thanks

atom501:
I wasn't sure which was the better solution. I ran the app after the updated model list and and it works now.

Many thanks for the app :)

PR #5866, opened 2026-08-08
Add HT7407 to the OLED list

Coxxs:
https://www.asus.com/laptops/for-creators/proart/proart-pz14-ht7407/

(Flicker-free Dimming is supported after the patch)
image


seerge:
@Coxxs hello, thanks

Since my existing registry checking fallback did not auto-catch your device, i started digging deeper into "why"

And I have found a more robust way to detect [OLED based on actual panel ID instead,](https://github.com/seerge/g-helper/pull/5867)

Can you try this build and check if it detects your device as OLED from start ? Thanks

[GHelper.zip](https://github.com/user-attachments/files/30855143/GHelper.zip)

Coxxs:
Thanks! I tried this branch (https://github.com/seerge/g-helper/tree/oled-detection), but it doesn't seem to detect the OLED on my device. Could this be because my device is using Qualcomm X2 Elite?

image

image

```
Panel ID: 00000000 OLED: False
```

seerge:
@Coxxs ok, apparently that endpoint either does not exist or don't report anything in your case, it's not about CPU tho

Can you open `Extra` settings in the app, click on 🔍 icon and upload scan.txt here ?

Coxxs:
[scan.txt](https://github.com/user-attachments/files/30857139/scan.txt)

seerge:
@Coxxs thanks! I don't see a panel ID here either, so likely it's not reported at all. And that would explain why Asus service did not mark device as OLED in a first place in windows registry that app checks as a fallback.

I guess it's indeed easier to hardcode your model name for now then :)

PR #5917, opened 2026-08-19
Overlay settings screen

seerge:
image

[GHelper.zip](https://github.com/user-attachments/files/31217121/GHelper.zip)


hongkhang194:
This is amazing! Would it be possible to add a 'clock' function to the overlay? It would be really nice to see the time so we don't get carried away while gaming.

gigantf15:
Hi @seerge 

Will u add an option about CPU, GPU and VRAM clock speeds in this settings tab?

PR #5918, opened 2026-08-19
feat: release workflow for standard and standalone .NET EXEs

cbossaers:
This is my first PR in a public repo so I apologize in advance if anything was done wrong.

This PR attempts to add an additional step in the release workflow, by building an additional executable that embeds .NET 10 in order to circumvent the need to install the tool if the user does not want/need to install it separately, or is for some reason restricted to doing so (i.e. managed devices).

This change does not remove the existing (now renamed as GHelper-standard) version that does not package .NET and requires separate installation.

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: windows-2022
     steps:
     - uses: actions/checkout@v7
+
     - name: Setup dotnet
       uses: actions/setup-dotnet@v5
       with:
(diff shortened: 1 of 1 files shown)


cbossaers:
Could I get some comment on this @seerge ? Thanks

seerge:
@cbossaers hello

I don't see point including it :) It makes exe HUGE, killing whole purpose of the app.

If you have some exotic usecase, you can do it for yourself of course

Issue #5822, opened 2026-08-01
Add native ROG Nebula ELMB toggle for 2026 ROG Strix SCAR 18 (G835LX)


synredge:
### Rules

- [x] I made myself familiar with the <a href='https://github.com/seerge/g-helper?tab=readme-ov-file'>Readme</a>, <a href='https://github.com/seerge/g-helper/wiki/FAQ'>FAQ</a> and <a href='https://github.com/seerge/g-helper/wiki/Troubleshooting'>Troubleshooting</a>.
- [x] I understand that, if insufficient information will be provided, my issue will be closed without an answer.

### Is your feature request related to a problem? Please describe

Yes.

The 2026 ROG Strix SCAR 18 (G835LX) includes a new Nebula ELMB display technology, but there is currently no way to enable or disable it from G-Helper.

Owners of this laptop currently have to keep Armoury Crate installed (or reinstall it)  to access this particular feature.

### Describe the solution you'd like

I'd like G-Helper to expose a native toggle for ROG Nebula ELMB setting, similar to how it already exposes other display-related settings.

### Describe alternatives you've considered

The only current alternative is installing Armoury Crate solely to toggle ELMB, which is not an ideal solution for those of us that want to continue using G-Helper.

### Device and Model

Asus SCAR 18 G835LX

### Additional information.

I'd be happy to help test experimental builds or collect logs if needed.

seerge:
@synredge hello,

Can you upload scan.txt as explained here https://github.com/seerge/g-helper/issues/38  ? then toggle ELMB to the opposite in armoury and upload scan again ?

Thanks

seerge:
@synredge any updates here ? :)

synredge:
@seerge Hi! Sorry for the delayed reply. Right after I posted this issue I left on a trip, so I’ve been away from my laptop.

I should be back soon and can definitely grab the scans for you.

One quick question before I do: is Armoury Crate required to toggle ELMB for this test, or is there another way to flip the setting that would produce the same before/after scans? I did a completely clean Windows install and have been trying to avoid installing Armoury Crate if possible.

seerge:
@synredge yes, you need to toggle in AC so I can see which endpoint is responsible for the toggle (likely it works in a very similar way to all other toggles like mini-led toogle or anything else works)

seerge:
@synredge any updates here ?

subi10:
this is my asus strix scar 18 2026 with ELMB off and on

[...]

PR #5962, opened 2026-08-31
ELMB support https://github.com/seerge/g-helper/issues/5822

seerge:


xxmustafacooTR:
ELMB currently only supported on 2026 model, LXG one.

